PR #9, opened 2020-03-12
Update notes re: Montana update time protocol

jdmaresco:


webmasterkai:
Thanks for this! I've updated the information in the “Notes About the Data” column of the "States" sheet.

PR #13, opened 2020-03-12
Initial style sketch for the State Notes page

beep:
Hello all! In this branch, I’ve got a speculative sketch of some styles for the State Notes page. I would really, really like any feedback y‘all might have to offer.

Here’s a short screen recording: https://www.dropbox.com/s/y51bhbnecy3kwc1/state%20notes.mov?dl=0

And here are some quick screenshots:

Screen Shot 2020-03-12 at 5 05 15 PM
Screen Shot 2020-03-12 at 5 05 59 PM

I’m coming at this content relatively cold, so any feedback/notes would be really welcome. Specific things I’m wondering about:

- Does a grid/card layout feel appropriate at wider screens?
- Does positioning the notes underneath each state’s table feel appropriate?

Let me know how this feels to you! Happy to take another crack at this okay thank youuuuuu

PR #20, opened 2020-03-13
Slight tweaks to states page

beep:
Did a quick tightening-up on yesterday’s states page work (#13):

- Pulled the Twitter icon closer to the state name, so that these tables don’t look like things Twitter generated. Because they are not.
- Refactored the table styles a bit (https://github.com/COVID19Tracking/website/commit/38769b3c3340b6d97006a80278824d703da8126d#diff-21411f591eb93a949eb6f1f8fd79ad70R104) so there’s a common `table` component that’s centrally accessible. (@webmasterkai, I think your summary table still looks 💖great💖, but please let me know if I donked anything up.)
- Bumped up the size on the `.updates` block, just to make it look a little more…page level-y yes this is a word now

Here’s a before & after GIF!

states

Would love to hear folks’ thoughts.

In addition to the above, I tweaked the BrowserSync config _very slightly_ in https://github.com/COVID19Tracking/website/commit/8b8a1b91caab409612d60b30c0e0ad837737cd67. @Wilto (or anyone else), does that seem okay to you? Happy to revert if not.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -4,6 +4,11 @@ const pluginRss = require("@11ty/eleventy-plugin-rss");
 module.exports = function(eleventyConfig) {
 	const CleanCSS = require('clean-css');
 
+	eleventyConfig.setBrowserSyncConfig({
+		ghostMode: false,
+		open: true
+	});
+
 	eleventyConfig.addPlugin(pluginRss);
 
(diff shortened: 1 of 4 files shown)


webmasterkai:
Looks good to me

PR #34, opened 2020-03-16
Configuring eslint and prettier

ccheever:
Setting up ESLint and Prettier for this repo

At @webmasterkai's suggestion, we are mostly using the Airbnb JS Style guide
except tweaking it to use semicolons as little as possible.

I set this up using this script: https://github.com/paulolramos/eslint-prettier-airbnb-react in the hopes that it would make sure that eslint and prettier
would play nicely together. I manually edited the configuration
files (`.pretterrc.js` and `.eslintrc.js`) to set semi to never/false.

I didn't run prettier or ESLint autofix over all the files in the repository
as part of this change. That could be done as separate commit if someone
wants to do it.

I also didn't setup any commit hooks or other enforcement mechanisms for
these policies. If these seem to be working well and we want to enforce
them more strongly later, we could add that. In the short term, it seems
like those could risk blocking us from pushing or editing the website and
that would be worse than having inconsistent style or formatting.

You can setup prettier in VSCode by just installing the VSCode extension and then (on Mac) pressing `Option+Shift+F`. It is also possible and common to configure VSCode to format on save.

In most of my own projects, I rely on prettier and haven't used eslint in a while, so I am more confident that it is correctly configured than I am that the eslint configuration is correct.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,29 @@
+module.exports = {
+  extends: ['airbnb', 'plugin:prettier/recommended', 'prettier/react'],
+  env: {
+    browser: true,
+    commonjs: true,
+    es6: true,
+    jest: true,
+    node: true,
+  },
+  rules: {
+    semi: 'never',
+    'jsx-a11y/href-no-hash': ['off'],
+    'react/jsx-filename-extension': ['warn', { extensions: ['.js', '.jsx'] }],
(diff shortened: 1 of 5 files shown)


webmasterkai:
I care more about having _something_ more than the details of what that something is. If there are strong feelings that this was the wrong choice of style choices we can change it.

Issue #21, opened 2020-03-13
Clarification on /states/daily counts


plmrry:
Are the values in `/api/states/daily` _cumulative_ or _new tests per day_?

I assumed they were cumulative, but noticed that the total value decreases over time for certain states (e.g. New York).

kissane:
@plmrry Can you drop this into https://github.com/COVID19Tracking/issues/issues please? There are data people wathcing those, this is for getting the website launched.

careeningspace:
They should be cumulative, but there are some data points that may have been revised do to a change in reporting methods. Do you have examples?

kissane:
So what happened in the example you emailed in, @plmrry—which, thanks!—is that NY stopped reporting two classes of numbers (negatives and pendings), and we didn't carry over the old numbers. Normally we DO carry over even old negatives and pendings, and we can look into that one.

hammer:
I can fix up the NY data to carry forward the negatives. I did the same for MD where the negatives weren't carried forward for a few days. Pending I just let go blank when they stop reporting.

rypan:
I've gotten the same question about our spreadsheet...if data on `US Daily 4p ET` is "daily" or "cumulative". We should clarify by updating the column title? (Or will that break things...for people who are scraping us?)

kissane:
I would be very much in favor of clarifying somewhere in the sheet and will leave it to @hammer to decide how to do it without breaking things

webmasterkai:
Has this been fixed? I just got an email from WSJ asking about it.

hammer:
I think documenting in the spreadsheet is hard so it's best to document on the website. We've successfully buried the spreadsheet link at this point so if you end up there without seeing any of our documentation and you miss the README tab that links to the website I'm sorry.

@rypan for your particular question it's cumulative.

hammer:
Another question I get regularly: what do update time and check time mean on the current tab, and how are the daily tabs computed?

I suggest we put this information on the `/data` for now.

kissane:
I'm bringing in someone from Politico with data questions

bzjin:
Sometimes the negative cases are null, which brings the total down day by day. 

For example, in Louisiana, `negative` cases are `37` on March 11 and March 13, but `null` on March 12. This makes the `total `cases 51, 14, then 51, which I believe is in correct.

webmasterkai:
Another example of negatives in NY the 11th and 12th.
[...]

PR #41, opened 2020-03-17
Document cumulative nature of data

jeremiak:
There were a few questions, albeit because of a data mistake, regarding if the data was cumulative or per day. We can fix the data mistake but let's also have the answers in our docs.

Refs #21


metasj:
Aza said today "any static count of cases should include a sparkline of the trend, or it's not really informative" -- anything we can do to indicate at least the rate of change would be great + could help address this visually.

PR #42, opened 2020-03-17
[US Daily Data] Fix date formatting

ccheever:
This makes the dates more readable and more consistent with other dates
on the site.

I also included the 3 letter abbrev for day of the week at the end of the
date since I think that is what I would want to see (weekends are sometimes
anomalous, sometimes easier to think in terms of days of the week when
thinking about the recent past)

This can be changed by adjusting the format string here:
    +      return DateTime.fromFormat(dayStr + '', 'yyyyMMdd').toFormat(
    +        'dd LLL yyyy ccc',

Taking out the ` ccc` will remove the day of the week

New look is this:
image

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -1,66 +1,75 @@
-const { DateTime } = require("luxon");
-const pluginRss = require("@11ty/eleventy-plugin-rss");
+const { DateTime } = require('luxon')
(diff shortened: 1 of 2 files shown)


webmasterkai:
Thanks!

PR #46, opened 2020-03-17
Icons.

Wilto:
Before and after:
image

This includes icons for various platforms’ “add to home screen” arrangements.


runxc1:
where is the add to home screen configured?

PR #43, opened 2020-03-17
Refactor CSS and design iterations

mdo:
After a [request from Julie](https://twitter.com/joulee/status/1239755367556337665?s=20) for some design/CSS feedback, I volunteered some time to check out the project and see where I could be of use. Holled at @ccheever as well to get a brief lowdown on things. This PR is very open to any and all feedback, or outright closure if y'all aren't into it :). Just holler with any questions or comments.

Below is a summary of changes and some screenshots. My changes largely focused on under the hood updates, optimizing CSS, removing duplicated code, some minor Markdown changes, and adding clear navigation. Hope it helps!

Anyway, here we go!

## Summary

- Updated links to have clear blue color throughout. As I understand it, the goal is to keep things black and white, but the links didn't seem that easily discoverable, even with the underline.

- Added basic navigation to the header. I know there are links in the footer, however, these are in the footer—quickly jumping on mobile and even desktop without scrolling to the bottom would be an improvement. Current page style would be nice, but I'm new to nunjucks. On narrow viewports, the nav wraps to a new line.

- Made table styles all global, reducing specificity and duplication. I didn't want to push the table styles too far without direction first, so the style changes were kept pretty minimal Striped rows were added for quicker scanning (used `even` for now to avoid extra coloring on the two-row state tables).

  Also made all tables `text-align: right` for the time being, and added a `.text-left` utility class, due to the tables exclusively housing numerical data. This makes all the data easier to read.

- Added explicit button styles for the homepage button, and shortened the button text. These could be easily dropped for simpler styling if that's desired.

- Un-nested nearly all `.body` styles for lower specificity. `.body` is left intact for some `font-size` changes and list styling, but the rest is top level now.

- Removed nearly all of the unused, duplicated `_states.scss` CSS from `_usDaily.scss`. Could maybe remove more.

- Added some new Sass files to help split things out. Let me know if this goes too far, happy to consolidate.

## Possible todos

- [ ] Update tables (or just US Daily table?) to be responsive, using some `overflow` and `white-space` to make them scroll horizontally maybe?
- [ ] Nav current page highlighting
- [ ] Adjust some page sizing/spacing to bring tables higher up on the page
- [ ] Further CSS refactoring

## Screenshots

covid1

covid2

covid3

---

Thanks for being awesome and making useful, beautiful open source projects that help people! <3

PR #48, opened 2020-03-17
Adding in a .editorconfig

beep:
If we’re going to start taking in more external contributions (à la https://github.com/COVID19Tracking/website/pull/43), it’d be great to have an [.editorconfig](https://editorconfig.org/) that documents a few of our expectations around (just e.g.) whitespace.

…so that’s what this PR is offering hi

This is definitely a draft, so comments/feedback are welcome. And this can definitely evolve! Treating this as a starting point more than anything.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,3 +11,7 @@ indent_size = 2
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+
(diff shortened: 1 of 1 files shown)


webmasterkai:
I added a couple markdown exceptions. Looks great to me.

PR #51, opened 2020-03-17
Display total number of tests since yesterday

lonnylot:
This adds the total number of tests since yesterday to `/data`

### DIFF
--- a/_src/_templates/states.njk
+++ b/_src/_templates/states.njk
@@ -16,6 +16,7 @@ layout: base.njk
         <th scope="col">Pending</th>
         <th scope="col">Death</th>
         <th scope="col">Total</th>
+        <th scope="col">Total Since Yesterday</th.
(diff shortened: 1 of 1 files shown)


webmasterkai:
Ideally I think we make this calculation available higher up the chain. Within the API or maybe the sheet itself? I'll wait on others to give some input here.

kissane:
It's a good feature rec! But yeah, we should either have it in the data sources or in the forthcoming Viz area of the site so our site data and API data and Spreadsheet data all mirror each other. Thanks, @lonnylot, I'll send the idea to the whole team.

careeningspace:
We can set up additional sheets to provide additiona APIs which can also feed other pages
You ask and DE&P™ will provide (within reason)  

*Data Entry & Publishing, a division of the COVIDTracking.com project

PR #56, opened 2020-03-18
Add some recent press that uses project data

jeremiak:
These links got shared in Slack, adding them here too. Also adding logos of press outlets that have used the data.

Screen Shot 2020-03-18 at 5 28 52 PM


webmasterkai:
We might want to double-check that all of these got added to the spreadsheet https://docs.google.com/spreadsheets/d/1-lvGZ3NgVlda4EcF5t_AVFLnBqz-TOl4YZxYH_mJF_4/edit#gid=0

PR #68, opened 2020-03-20
markdownify state notes

schwartzadev:
Currently the state notes are rendered as text, not as markdown:

image

This PR should fix that.


olivierlacan:
Good catch!

PR #69, opened 2020-03-20
Differentiate Last updated from Last refreshed

olivierlacan:
image

We don't actually "update" any data when we refresh/pull from the API. Which means
the timestamp listed falsly suggest we update the data show on the States and
US Daily pages when `sheets.updated` changes.

I think "refresh" avoids that potentially confusing term but it does leave
some ambiguity which I think could either be solved with copy on the page or
using an info tooltip for people who care what "Last refreshed" means exactly.

By contrast, the `Last updated` we show for each state *is in fact* when that state last updated their data according to timestamps we report back. So I think it's important to distinguish these two uses of the term. 

image

### DIFF
--- a/_src/_templates/states.njk
+++ b/_src/_templates/states.njk
@@ -3,7 +3,7 @@ layout: base.njk
 ---
 
 {{ layoutContent | safe }}
-<p class="updated">Last updated: {{ sheets.updated }}</p>
+<p class="updated">Last refreshed: {{ sheets.updated }}</p>
 
 <div class="us-totals">
   <table>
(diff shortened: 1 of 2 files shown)


beep:
Hey @olivierlacan, thank you so much for catching this! We had a brief chat in Slack: would you mind using the text `Last synced with our spreadsheet`?

olivierlacan:
@beep Looks like @webmasterkai tackled it. Yay for PRs with commits open. 🤗

beep:
Eep, yep, never mind! Thanks, @webmasterkai.

PR #71, opened 2020-03-20
Add page listing visualizations built using our data

### LINKED ISSUE #16: Add visualization gallery
Could include things like Willem’s dope Tableau dashboard and graphics that use our data like this one:

olivierlacan:
Fixes #16

Added as a subtle new hyperlink to the copy at the top of the Data page: 

image

The page itself is a simple list for now:
image

This is a first draft without pictures but that's an exponentially hard
problem so I'm not sure it's worth me taking a bunch of time to take screenshots
unless we can find some library that can simply unfurl some OpenGraph image tags
or create screenshots semi-automatically.

There's minor overlap with the work in #56 but this page focuses more on the
value added by the visualizations, not so much on the press coverage of this
project so I think they both deserve to exist.

I didn't add a nav item for it (yet), because I feel like we're getting crowded
down there and this is an early version of this. But at least we can point people
to it right away.

I've also spent the time to find and hyperlink the authors of these because
there's some damn fine data viz work being done and these people (not just their
employers in the case of press) deserve the hat tips too.


webmasterkai:
Added the page so it's viewable to the team, but removed the link until we get broader approval.

https://covidtracking.com/visualizations/

A link `[explore visualizations built using our data](/visualizations)` will need to be added somewhere so people can get to it. The `/data/` page is probably the most obvious.

PR #72, opened 2020-03-20
Use relative links when appropriate

olivierlacan:
There's no need for us to hardcode the domain for any path that falls under
covidtracker.com inside of our Markdown. It adds visual clutter for no good
reason.

The only slight advantage on the API page is that the API links work on
localhost development if they're pointing to covidtracker.com and they won't
work with this change since you might not be running the API and routing it
appropriately.

Still, I don't think maybe people hit the API endpoints from their local
machine.


webmasterkai:
Initially the API was on a different domain so it needed the full URL. I agree with you here. Full URL no longer needed. Thanks!

PR #75, opened 2020-03-20
Minor accessibility tweaks

kevee:
This PR fixes some minor issues when I did an accessibility audit:

- [x] Remove [empty UL](https://github.com/COVID19Tracking/website/blob/master/_src/_templates/us-daily.njk#L9) around US daily table
- [x] Remove empty link after first sentence on the [Data page](https://github.com/COVID19Tracking/website/blob/master/_src/data.md).
- [x] Individual state page links to Twitter should read "[state name] on Twitter"

Also, this came up, but I think is fixed on #73 
- [ ] The Data page has 50 links that all say "Best current data source", which is confusing in a link selection rotor. Add "in [state name]" to the link. If we want to not have a long link name, you can wrap "in [state name]" in an off-screen readable element, or use `aria-label`

### DIFF
--- a/_src/_templates/state.njk
+++ b/_src/_templates/state.njk
@@ -8,7 +8,7 @@ layout: base.njk
     <ul class="state-links">
         {% if state.twitter %}
             <li>
-                <a href="https://twitter.com/{{ state.twitter }}">Twitter</a>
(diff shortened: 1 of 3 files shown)


beep:
_BEAUTIFUL_. Thank you so much, @kevee!

(Also, yes, I can confirm that the duplicated link text is fixed in #73!)

PR #80, opened 2020-03-20
add twitter svg icon, replace twitter icon on states page

schwartzadev:
This icon is from Twitter's Brand Resources page -- I modified it by changing the color of the icon from white to black and removing the background


schwartzadev:
I would like to add the Twitter icon to the individual state pages as well, but I'm not sure where to put it, from a design perspective... Any suggestions?

webmasterkai:
Can you double-check the "Best current data source" that got added back in here?

schwartzadev:
⚠️ WIP

schwartzadev:
There it is! Sorry about that.

image

PR #84, opened 2020-03-20
State page twitter icons

schwartzadev:
Adds Twitter icons to the state pages instead of the Twitter link in text:

New:
image

Old:
image

### DIFF
--- a/_src/_assets/css/_state.scss
+++ b/_src/_assets/css/_state.scss
@@ -18,6 +18,8 @@
     }
     a {
       align-self: flex-end;
+      text-align: right;
+      flex-basis: 31px; /* this adds 5px to the left side */
     }
   }
 }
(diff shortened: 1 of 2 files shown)


schwartzadev:
Hey it looks like the CI checks are hanging...if you click through to the Netlify details, it looks like they're done.

lucasgonze:
@schwartzadev did it look to like the CI timed out some tests that did eventually pass? In that case we should look into fixing the tests.

schwartzadev:
I'm not sure what you mean... For reference, the Netifly log indicates that everything worked successfully in ~51 seconds, which seems typical. I think the issue might be Netlify talking to the CI on Github: Netlify says the checks are completed successfully [on its site](https://app.netlify.com/sites/upbeat-lovelace-3e9fff/deploys/5e75084474cc510007ef553d), but Github isn't showing the same.

lucasgonze:
yes, we understand it each other. It seems to me this is mergeable, but there is an issue with where Github is failing tests because it's not waiting long enough. Maybe there is an admin setting for that.

PR #94, opened 2020-03-21
States data quality grade

goleary:
This one looks good:
`/data`:
image

This one could use a little love:
`/data/states`:
image

description of the grades (mostly pilfered from the twitter status) here:
image


kevee:
This looks great! The only thing I'm concerned about is just linking to a twitter status - making that information canon somewhere in the actual site is I think preferrable.

goleary:
@kevee agreed, I wonder what the best place for that is...  /about-tracker?

I'll put something together

under/linked from this section:
image


Also I can't help but feel slightly bad about duplication across multiple files in this PR...I haven't worked with 11ty/nunjucks before.  perhaps there is some way to make this thing a component referenced from both `state.njk` & `states.njk`

_src/about-tracker.md
```diff
@@ -19,6 +19,25 @@ We get the bulk of our data directly from state public health authorities, so we
...
+A number score is tallied based on 4 simple components:
+
+1. +1 for reporting positives reliably (only NV fails)
+2. +1 for reporting negatives sometimes
+3. +1 for reporting negatives reliably
+4. +1 for reporting all commercial tests
+
+The total score for each state corresponds to a letter grade:
+
+4: A, 3: B, 2: C, 1: D
```
schwartzadev: I would present this as a table to increase readability -- maybe try:

```
| Score | Grade |
|-------|-------|
| 4     | A     |
| 3     | B     |
| 2     | C     |
| 1     | D     |
```

olivierlacan: Seconded.

webmasterkai:
Note that the grade is now in two places:
Screen Shot 2020-03-21 at 5 00 51 PM

goleary:
> Note that the grade is now in two places:
> Screen Shot 2020-03-21 at 5 00 51 PM
> Maybe one of those should be removed?

think it got added back in a merge.  I just removed the original one.

_src/about-tracker.md
goleary: Just modified it however it seems to waste quite a bit of space:
image

schwartzadev:
Looks like the double grade came back -- could you please remove the smaller one?

image

PR #99, opened 2020-03-21
Handle missing grades

schwartzadev:
This handles missing grades by labeling the grade as "N/A" if it is not available.

----

Current (states page):
image

Updated (states page):
image

----
Current (state page):
image

Updated (state page):
image


schwartzadev:
Not sure why this is happening on my PRs, but it seems like Netlify is having trouble talking to Githhub. I've never worked with Netifly before, but hopefully someone who is familiar with the service could take a look?

PR #100, opened 2020-03-21
[DO NOT MERGE] - First up for the dataviz dashboard

jeremiak:
Screen Shot 2020-03-21 at 7 26 29 PM

Adds a new /dashboard route and experiments with the `britechart.js` library and a quick bar chart. Not perfect or even done, just pushing up to get feedback from the rest of the team.


kevee:
Hi, @jeremiak, this is looking great. I think the only piece of feedback I have is on the accessibility front. Easy one is the SVGs should have a readable `title` or `aria-label` attribute.

Harder one (this is always a challenge with charting) is making the data accessible. I'm having to borrow a computer in my time of quarantine, so I'm not able to test with JAWS, but Voiceover is reading the chart just using the X axis then the Y, not reporting values.

There's two approaches, one is to [use ARIA roles to specifically label ticks](https://www.w3.org/wiki/SVG_Accessibility/ARIA_roles_for_charts) in the chart. In looking at britechart it doesn't look like there's an easy way to do that with their native API.

The other option is to append a simple table (table data is really easy to navigate with a screen reader). Just add the class `.a11y-only` to the table so it's not visible. Then explain that there's a table after the chart in the SVG's `title` or `aria-label`.

jeremiak:
That's really awesome, actionable feedback @kevee - thanks!!

jeremiak:
Closing in favor of #136

PR #106, opened 2020-03-22
Tables redesign (Closes #103, partially closes #101)

### LINKED ISSUE #101: Design updates for mobile
I've noticed a few rough edges when viewing the website on my iPhone.

### Home
- [ ] "US Historical Data" and "Most Recent Data" in the header wrap onto two lines.
- [x] The "In the press" logos stretch the width but the list below wraps much earlier, causing the logos to have a big overhang.
- [x] Footer links stack vertically quite a bit. Maybe fine but looks a bit odd.

### Most recent data
- [x] "Pos + neg" title in US totals table wraps
- [x] Below each state the "Grade - Last updated - Last checked" display wraps. Maybe stack?

I'll stop here but when we have resources this could use a pass.

### LINKED ISSUE #103: Hospitalization column is causing visual collision
On: https://covidtracking.com/data/

image

We probably need to rework the way we set up the grid since it's now overflowing.

schwartzadev:
Edit: no longer a WIP

This closes #103 and may address #101 .

Edit: this closes 2/5 of #101 

More description to come soon.

### DIFF
--- a/_src/_assets/css/_states.scss
+++ b/_src/_assets/css/_states.scss
@@ -1,11 +1,9 @@
 // Specificity hack is specific
 .us-totals {
   max-width: 60rem;
+  margin-bottom: 3em;
 
-  table {
-    margin-bottom: 4em;
-  }
-  caption {
+  .state-table-caption {
     padding-left: 0;
     text-align: left;
   }
--- a/_src/_assets/css/_tables.scss
+++ b/_src/_assets/css/_tables.scss
@@ -8,7 +8,7 @@ table {
   }
 }
 
-caption,
+.state-table-caption,
 th,
 td {
   padding: 0.25em;
@@ -37,6 +37,7 @@ tbody tr:nth-child(even) {
   display: flex;
   flex-wrap: wrap;
   flex-direction: column;
+
   @media (min-width: $viewport-md) {
     flex-direction: row;
   }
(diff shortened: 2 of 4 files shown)


schwartzadev:
This PR takes the existing table structure for tables on the `/data/` page (aka the `states` page) and moves it into a flexbox structure.

The major change is that the table structure can be modified based on the device width (i.e. for mobile) via media queries.

I'll mention some minor changes below.

----

Mobile
----
**New state table:**
image

**New us-total table:**
image

**Old state and us-total tables:**
image


Desktop
----
The appearance on desktop remains largely the same:
(Note that this PR does add "N/A" for null state values)

**Old state table:**
image


**New state table:**
image


**Old us-total:**
image

**New us-total:**
image


Minor Changes
----
As promised, I'll discuss these here. As noted in #103 , some table previously extended out of their intended layout. This PR changes that using `flex-grow`. The behavior is such that, if a table is outside of bounds, it will wrap below the rest of the content.

This currently happens with California since its values are wider than most states:
image

This can also happen at certain device widths (this is 907px wide):
image

To reduce the likelihood of this wrapping, we could reduce the width of the column headers (namely `Hospitalizations`). Regardless, I think this behavior is more desirable than the collisions @olivierlacan identified in #103 .

olivierlacan:
Hold off from merging, I've made a boo boo when solving the conflict.

Issue #83, opened 2020-03-20
Header anchors to link to specific section


olivierlacan:
I found myself wanting to link people to the Best Practices doc from the homepage and since the page overflows the default viewport height on most screens I realized it wasn't easy to link to a specific section based on a header anchor link (which I find extremely useful).

Example: 

```diff
-<h2>Best Practices for Data Release</h2>
+<h2 id="best-practices-for-data-release">Best Practices for Data Release</h2>
```

Then anyone can easily link to `https://covidtracking.com/#best-practices-for-data-release` and get straight to the relevant section. We can also style the headers to display a `¶` symbol next the header when hovered so these permalinks are discoverable.

I have a proof of concept that simply adds [markdown-it-anchor](https://github.com/algolia/markdown-it-anchor) and replaces the default 11ty Markdown generator so that we can add that plugin and process all headers with it (it adds an anchor wrapping header tags and we can customize which header levels get it).

I'll post the PR in a bit for discussion if we think this is useful.

webmasterkai:
I'll add a redirect for the google doc https://covidtracking.com/best-practices

olivierlacan:
That's swell too. Think anchor links are worth it anyway?

kissane:
Where did this lovely idea end up?

olivierlacan:
@kissane Forgotten a bit by this grueling past week and the churn I worried about on the website front. I'm sort of waiting for the Gatsby stuff to get merged in before attempting to do this since the implementation I had started was heavily dependent on the Markdown parser used by 11ty and its configuration.

kevee:
Closing, as markdown has links in gatsbyjs branch

PR #110, opened 2020-03-22
shrink table width

schwartzadev:
I don't think there is a way to do this in MD, so I ended up using HTML here, with inline styles (not great, I know).

See #94 -- this table currently takes up a lot of space.

Here's the updated version:

image

I won't merge this yet since I want someone else to be on board with this implementation of an HTML table in the middle of this Markdown file. If anyone has a cleaner way to achieve this, I am all ears.

### DIFF
--- a/_src/about-tracker.md
+++ b/_src/about-tracker.md
@@ -32,12 +32,33 @@ A number score is tallied based on 4 simple components:
 
 The total score for each state corresponds to a letter grade:
 
-| Score | Grade |
-|-------|-------|
-| 4     | A     |
-| 3     | B     |
-| 2     | C     |
-| 1     | D     |
+
+<table style="width: 200px;">
+  <thead>
+    <tr>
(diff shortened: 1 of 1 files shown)


goleary:
well it definitely looks better...
I'm also not a markdown wizard and had to stick some html above to be able to link to a section of this page.

schwartzadev:
Hmm yeah... About the linking -- @olivierlacan opened #83 , which would add anchors to every header, a little while ago but it looks like it got lost in the shuffle.

The [markdown-it demo](https://markdown-it.github.io/) doesn't seem to support this type of behavior but does seem to be very extendible -- maybe that's an option?

olivierlacan:
> The [markdown-it demo](https://markdown-it.github.io/) doesn't seem to support this type of behavior but does seem to be very extendible -- maybe that's an option?

I have a branch I'll push up

PR #113, opened 2020-03-22
Added Jest tests for API, build steps

kevee:
This adds two new steps to the build process (also available as npm scripts):

`test-api` - Checks that the API endpoints are up, are returning valid JSON, and objects match expected values.
`test-build` - Checks that homepage, data page, and individual state pages are built.

These were added to the `prod` script to make sure everything is OK before netlify publishes.

### DIFF
--- a/_src/__tests__/api.js
+++ b/_src/__tests__/api.js
@@ -1,0 +1,136 @@
+const fetch = require('node-fetch')
+
+describe('Data API', () => {
+  it('returns state data', done => {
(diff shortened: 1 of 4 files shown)


kevee:
After fiddling with netlify a bit, this is now working. It first tests to see if the API is running, and then runs grunt/eleventy, then tests that the site built with at least California. 

Long term we could use the API again to test all states exist, etc, but this should help at least make sure that **some** API data was ingested and built.

PR #114, opened 2020-03-22
Added state navigation

bgamrat:
Added state navigation to right side


schwartzadev:
Thanks for the addition!

If you don't mind, I'm going to make some minor changes to the formatting and the mobile behavior.

Right now, this nav takes up a lot of valuable real estate on mobile devices so I'll try to find a fix for that.

image

hammer:
Could we do this with a map too or instead?

schwartzadev:
Ooh that could be cool...Do you mind if I do the abbreviations in the meantime and then look into a map?

schwartzadev:
Desktop
image

Mobile
image


I really don't love the arrow placement -- from a UX side, it's not clear what it means. Instead I would add a `(return to top)` link where I placed the two grey lines.

image

@bgamrat Do you want to do that? If not, I can take care of it 👍

bgamrat:
No worries on making changes - it’s your stuff

Thank you for asking

schwartzadev:
image

image

All done!

bgamrat:
Nice!

PR #117, opened 2020-03-22
Simplify and fix action to rebuild site twice per hour

jakejarvis:
Hi there! I noticed that the GitHub Action you're using is doing some funky stuff right now with four updates queued up so I'm hoping these changes/simplifications will make it more reliable.

* Switched to calling `curl` directly on the virtual machine instead of building a whole Docker image to do so. This should speed things up and remove a point of failure that actions sometimes get hung up on, which is probably what's happening now.
* Fixed the cron syntax to run at the 29th and 59th minute of every hour (I assume the 29 was because the site takes about a minute to build?) Previously it was running three times per hour: at the 29th minute, and the 58th minute, and at the top of the hour (so two minutes apart). Cron jobs are such a headache sometimes...
  * Before: [https://crontab.guru/#*/29_*_*_*_*](https://crontab.guru/#*/29_*_*_*_*)
  * After: [https://crontab.guru/#29,59_*_*_*_*](https://crontab.guru/#29,59_*_*_*_*)

Thanks for all the great work everyone is putting into this! ❤️


webmasterkai:
Thank you! This is great.

olivierlacan:
@jakejarvis Nice to see you here. 👋

Issue #109, opened 2020-03-22
Double border on tables


schwartzadev:
This was my doing and should be the literal last priority.

----

The horizontal top and bottom borders are half as thick as the vertical ones. They should be twice as thick as they are now. (They should match the thickness of the vertical right and left borders.)

![image](https://user-images.githubusercontent.com/18607205/77242930-7a240180-6bc9-11ea-8af2-e15039d82c53.png)

Here's a close-up:

![image](https://user-images.githubusercontent.com/18607205/77242929-75f7e400-6bc9-11ea-97fc-82a02449af2d.png)

The top and bottom borders should be just as thick the rest of the table.

A fun game of hide-and-go-seek in CSS if anyone is up for it :)

beep:
👋 Hey @schwartzadev, just checking: will this issue still be relevant once v2 launches?

schwartzadev:
I don't think so -- thanks for following up 👍

Issue #115, opened 2020-03-22
Flexbox collapses with empty values


schwartzadev:
This should be a simple fix -- just need to go into the template and add whitespace values as defaults when the table values are missing.

Alternatively, this could be solved with `flex-grow`.

![image](https://user-images.githubusercontent.com/18607205/77255843-108b0e00-6c30-11ea-8362-bad3bae5c294.png)

schwartzadev:
closed by #120

PR #120, opened 2020-03-22
Fix table collapsing with empty values

bhays:
CSS fixes for #115 and #109.


olivierlacan:
Hey @bhays can you name your PR more descriptively to explain what it changes please?

schwartzadev:
I don't think this closes #109 but we should merge this anyways
image

#115 looks good -- thanks!

schwartzadev:
I don't think I was clear about #109 -- the issue is that the borders are different widths, not that they are different colors. I reverted the color changes.

Thanks for fixing the table!

bhays:
Can do, sorry about that.

The fixes include the `border-color` for the table header cells and adding `flex-grow` on `.state-flex-data` to avoid empty cells collapsing.

PR #128, opened 2020-03-23
Report positive test rate for states

pcj:
I was interested to see how testing outcomes differ by states and created this PR.

Attached is screenshot on the output:

positive-test-rate

### DIFF
--- a/_src/_assets/css/_states.scss
+++ b/_src/_assets/css/_states.scss
@@ -129,3 +129,12 @@
     margin-left: 5px;
   }
 }
+
+.statistics {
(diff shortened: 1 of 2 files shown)


kevee:
Hi, @pcj, thanks for the PR! I'm checking in with a few other folks on this number and will get back to you soon.

kevee:
So after checking in with our team, it was decided that we are not going to include this metric until and if state data quality improves. 

That's not to say that we don't value your contribution! Please share any other ideas you have about improving the site.

pcj:
OK, understood.  Reach out if you need specific help in other areas.

PR #132, opened 2020-03-23
Add lighthouse audit to PRs

kevee:
Adds an action to do a Lighthouse audit using [treosh/lighthouse-ci-action](treosh/lighthouse-ci-action). This PR also includes a `budget.json` file to define [page size budgets](https://web.dev/use-lighthouse-for-performance-budgets/). 

I'm including this so we can start getting baseline performance audits. We can refine budgets and pass/fail assertions like first paint for critical paths in the future.

### DIFF
--- a/.github/budget.json
+++ b/.github/budget.json
@@ -4,7 +4,7 @@
     "resourceSizes": [
       {
         "resourceType": "total",
-        "budget": 200
+        "budget": 1000
       }
     ]
   }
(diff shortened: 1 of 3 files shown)


kissane:
I don't know enough to bless this, @webmasterkai are we good to go here?

kevee:
Let's leave this open for now, Netlify hung up on the content rules and I want to make sure it works before we merge.

PR #135, opened 2020-03-23
More padding for states nav near medium viewport

### LINKED ISSUE #133: State shortcuts show up on top of the UI in some narrow viewport sizes
image

This happens with `#main` width set to 1024 pixels which I believe is the same as the body width.

kevee:
Fixes #133


kevee:
Reviewed in issue, merging.

PR #142, opened 2020-03-23
[⚠WIP] Redo nav

schwartzadev:
⚠ WIP

This updates the nav with work from @kissane @kevee .

This should just be an interim update until the redesign is rolled out.

Right now, this PR is **not** ready to be pushed -- many of the header and footer links do not yet have the proper `href` elements to actually work as links


webmasterkai:
I haven't looked at v2 yet but this should probably be in a data file (yaml maybe?) or otherwise fetched from the page md files somehow.

schwartzadev:
afaik this will be in YAML!

Issue #127, opened 2020-03-23
Sort shortcut state abbreviations alphabetically 


jdmaresco:
I think the abbreviations should be sorted alphabetically based on the abbreviation, not on the full state name. For a minute I thought Nevada was missing but then I realized NV was sorting before NH. 

![DE93ECB8-BF50-4FE6-86C2-0E0C335BF1BF](https://user-images.githubusercontent.com/1327983/77276682-a1460600-6c91-11ea-8f0b-20a3aafbfec0.png)

schwartzadev:
Nice, that's a good call. The [current code](https://github.com/COVID19Tracking/website/blob/master/_src/_templates/states.njk#L63) iterates using:

```js
{% for state in sheets.states %}
```
...which seems to be sorted alpha by state name. (Not state abbreviation.)

I'm not sure if sorting by abbreviation is something that we could do in the template...if not, we could expose the list of abbreviations in `sheets.js` and then consume it in the template like:

```js
{% for state in sheets.stateAbbreviations %}
```

Does anyone know if this templating can handle that type of sorting?

kevee:
Fixed per https://github.com/COVID19Tracking/website/commit/38b2838b58aab461fc196082618f437e153efc15

PR #143, opened 2020-03-23
#127, sort shortcut state abbreviations alphabetically

rootart:


kevee:
Looks great, @rootart, thanks for the help!

Issue #144, opened 2020-03-23
Fix missing icons for mobile 


olivierlacan:
![image](https://user-images.githubusercontent.com/65950/77367010-9fcb1b00-6d2f-11ea-8af6-5e31215cfedc.png)

I'd focus on the icons first, the rest we can probably tackle later.

Here they are: https://github.com/COVID19Tracking/website/blob/274b520564da7a171efafae59565cacaad96518e/_src/_templates/base.njk#L21-L24

bhays:
Happy to fix other missing assets as well, although I'm not quite sure on how they'd work with the current build setup.

PR #147, opened 2020-03-23
Add missing icons, run old icons through ImageOptim

bhays:
Add missing icons sizes from #144. Includes 152, 144, 120 and 114. Old icons were run through ImageOptim to reduce file size slightly.


schwartzadev:
This looks good to me -- minimization is :100: !

Thanks @bhays

PR #166, opened 2020-03-24
Goleary/dashboard map updates

goleary:
- new and improved legend
- removed ability to pan/zoom + controls
- ability to use d3 color schemes
- attribution for population data/shapefiles
- now using Albers US projection (from d3)

image


kevee:
This looks great! I added a PR (#168) to this branch to add a link to the accessible version of the data.

PR #168, opened 2020-03-24
Added a11y link to dashboard map

kevee:
Added a non-visible link to the data page after a map so users with screen readers can access the data.


goleary:
looks good, thanks

PR #169, opened 2020-03-24
Added states and territories to header

### LINKED ISSUE #98: "States" Refactor to "States and Territories"
This should be pretty simple: the Project currently always refers to states _and_ territories as simply "states"

For example, on the `/us-daily/` page, the table indicates that 56 states are tracked while this includes N. Mariana Islands, Guam, Washington D.C., and others:
image

I'm not sure how many instances there are where we use this phrasing, but I think it should be revised.

kevee:
Fixes #98


schwartzadev:
Looks good -- we should keep an eye out for other instances of this during the redesign.

Thanks!

PR #184, opened 2020-03-25
fix: Changed totals to positive + negative

### LINKED ISSUE #183: fix: Clarify test columns
Rename the "Totals" columns in US and state tables to "Total test results" and make sure it's clear that it is Positive + Negative

kevee:
Awaiting feedback!

Fixes #183


schwartzadev:
Thanks for doing this!

A couple things:

1. I believe that the tables will always wrap at this point:
image

I don't have any issue with this; I am just observing.

2. The in-template addition for each state doesn't seem to work. See CA above -- every state returns "N/A"

3. The U.S. table has "Positive + Negative" in the data cell while the states have "Positive + Negative" in the header cell. I'm not sure if this is intentional or not. (See US below and CA above)
image

This seems like an important PR for clarifying the data -- thanks for tackling it @kevee

kissane:
I would personally rather see the "positive+negative" in the header than the data cell

schwartzadev:
I think I see why the `Positive + Negative` info was not in the header...when I added it back, I get this lovely table:
image

Taking a deeper look now...

kissane:
Oh FUN thanks @schwartzadev. If we have to launch this ugly (with text in the data call) to save time, let's launch it ugly and fix it in the V2.

schwartzadev:
Merging this -- the less ugly option (in my opinion) is the current one with the text in the data (not the header) part of the table.

This should be irrelevant by the redesign soon 🤞

PR #188, opened 2020-03-25
changing press endpoint to conform to others and not hit the cache

goleary:


schwartzadev:
Oop this was my doing -- thanks Gabe!

PR #193, opened 2020-03-26
change state-nav bar to a hoverable dropdown menu at fixed position

swiperthefox:
Problem: Currently on the state data page, the state navigation bar will be out of view for most states. To choose another state, the user needs to go back to top,  then scroll through the state navigation list to find the target state.

Changes made in this pull request: 
1. States is shown in a grid (as the mobile version) to avoid the scroll, and it is hidden.
1. "jump to" label will be positioned as "fixed".
2. When the mouse hovers on the label, the state grid will appear.
3. the label text changed to "jump to state" because "jump to" feels incomplete without the state names following it.

Remaining issues: the "jump to" label is in the middle of page, does not fit well with other contents. Ideally, the label should be able "scroll-then-fix": scroll to the top first, then become fixed. But this requires extra JS code.

### DIFF
--- a/_src/_assets/css/_states.scss
+++ b/_src/_assets/css/_states.scss
@@ -17,9 +17,8 @@
 }
 
 .states-nav {
-  max-width: 6vw;
   padding-left: 5px;
-  position: absolute;
+  position: fixed;
   right: 1vw;
   top: 150px;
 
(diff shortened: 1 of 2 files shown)


schwartzadev:
Neat. Thanks for implementing this! I think we'll put this on hold for now since we're in the middle of [a redesign](https://github.com/COVID19Tracking/website/projects/1).

This is great food for thought for implementing state-by-state navigation, though. Thank you!

kevee:
Closing since we are launching the new site, but I'm noting this in another issue for the future.

PR #194, opened 2020-03-26
Start adding bubble map

jeremiak:
We should talk about styling and how we want this to work (like do we want it to have a scrubber that moves through time?) but I started converting our map away from leaflet since that's a thing we want to do anyway (plus it gives us way more visual styling options and a better projection)

Screen Shot 2020-03-25 at 11 09 09 PM


goleary:
work has already been merged in #230

PR #198, opened 2020-03-26
Add “Projects Built with Our Data” section

lachlanjc:
Hey friends! I’m not sure this is the right place on the website, so please let me know if you’d like it elsewhere, but I think it’d be great to have a list of other sites/projects that use data from here. I’ve added mine to start:

https://testing.predictcovid.com/

Thanks for all your hard work <3


schwartzadev:
Hey Lachlan, thanks for suggesting this -- I like this idea, but we would probably want a more complete list of projects to include.

@kissane -- I'm curious to hear your thoughts about this.

webmasterkai:
I'd suggesting adding it to the visualizations page. https://github.com/COVID19Tracking/website/blob/master/_src/visualizations.md

lachlanjc:
Ah, great! I’ll add it over there. Thanks.

I couldn’t find this page on the website—maybe we should link it in the footer?

PR #199, opened 2020-03-26
Data page, basic layout

kevee:
Added Typography.js so we have a baseline
Created a "Container" component
Renamed the header a bit because it used the class name "container"
Added a special "noContainer" to frontmatter so that pages like the homepage can control where containers appear
Added the container manually to homepage per above
Installed Reflexbox for simple flexboxing
Added basic state information to the /data page to mirror what's live.


schwartzadev:
Thanks Kevin!

PR #203, opened 2020-03-26
Refactor out area chart builder

jeremiak:
Not a complete implementation as the special case handling for states with super big numbers (like NY) isn't yet implemented but there's one less chart using britecharts, but I think if approved we can merge at this stage to keep it close to `dashboard`


goleary:
looks good, definitely cleans up dashboard-charts.js a good amount

PR #221, opened 2020-03-27
V2 summary tables design

schwartzadev:
Based on designs from @beep and co.

This work is in progress. @kevee -- please hold off on merging for now.

Edit: **many table values are missing since the API is not yet configured. Check todos for more detail.**

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -1,6 +1,7 @@
 const { createFilePath } = require(`gatsby-source-filesystem`)
 const path = require('path')
 const slugify = require('slugify')
+const objectHash = require('object-hash')
 
 exports.onCreateNode = ({ node, getNode, createNodeId, actions }) => {
   const { createNodeField, createNode } = actions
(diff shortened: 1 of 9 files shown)


schwartzadev:
This implements the `tablesaw` library for adjusting summary tables.

This has some performance implications on the client-side, namely that the `/data/` page has to render ~60 `tablesaw` tables, which takes 10-30 seconds after the first paint on my machine.

For me, this makes this page feel very clunky. I would urge others to pull this branch and check it out for themselves. We should probably discuss as a team before merging.

beep:
> For me, this makes this page feel very clunky. I would urge others to pull this branch and check it out for themselves. We should probably discuss as a team before merging.

I can’t get gatsby up and running, but if we’re seeing _10-30 seconds_ of lag, I’d be inclined to just overflow the tables for now. We can figure out a Fancy Adaptation Approach™ at some point after launch.

schwartzadev:
I'm going to remove the values that are missing from the API

schwartzadev:
Ok -- this looks good to go :+1: -- tables scroll on smaller viewports and behave like tables on larger ones.

PR #222, opened 2020-03-27
Add stay at home annotation

jeremiak:
Screen Shot 2020-03-27 at 4 05 10 PM

### DIFF
--- a/_src/_assets/js/d3-area-chart.js
+++ b/_src/_assets/js/d3-area-chart.js
@@ -22,6 +22,7 @@ function d3AreaChart({
     right: 55,
     bottom: 40,
   },
+  annotations = [],
 }) {
   const grouped = d3.nest()
     .key(function(d) {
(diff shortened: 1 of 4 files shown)


goleary:
Not sure if this is a result of this PR or something else, but the NY chart overflow is turned off and there are more states that are now overflowing
image

jeremiak:
That was part of the refactoring I did for the area chart. I'm working on a fix for that

_src/_assets/js/d3-area-chart.js
```diff
@@ -133,5 +134,21 @@ function d3AreaChart({
     .attr('opacity', 0.9)
     .attr('fill', fillArgument)
 
+  if (annotations) {
```
goleary: this will always be true unless someone explicitly passes `annotations:null` to the function.
perhaps better to check use `if(annotations.length)` or pass `null` as default value for `annotations`.

_src/_assets/js/dashboard-charts.js
```diff
@@ -192,7 +205,7 @@
       width: chart.node().clientWidth * .9,
       yMax: d3.max(transformedData, function(d) {
         return d.value
-      }),
```
goleary: are you using the prettier settings that are included with the project?
I autoformat before I commit code and it's sometimes reformatting your code
image

_src/_assets/js/dashboard-charts.js
```diff
@@ -427,7 +440,10 @@
         .append('div')
         .classed('chart', true)
         .classed('no-y-axis-domain', true)
+      const stayAtHomeOrder = stayAtHomeOrders[state.key]
+      const annotations = stayAtHomeOrder ? [{ date: parseDate(stayAtHomeOrder) }] : null
       const chart = d3AreaChart({
+        annotations,
```
goleary: ahh looks like you are passing `null` for annotations here, but not everywhere else this function is called.

_src/_assets/js/d3-area-chart.js
goleary: you could also use `if(annotations && annotations.length)` to prevent this block from being run with an empty array.  Don't think it matters that much because it does nothing if passed an empty array, just for code cleanliness.

jeremiak:
These are all great points, I'll integrate this feedback in my next PR. Merging this so we can maybe take screenshots

PR #228, opened 2020-03-28
[WIP] Gatsbyjs header tab

kevee:
Adds a first-pass to the newly designed headers. Uses a 'navigation' context and YAML files to define the sub-page tabs on the top.

Also, a mobile design & footers. Not sure if we want more navigation options in the footer.

Preview here: https://5e7faa687a200a0008f0a459--goofy-tesla-43452a.netlify.com/


schwartzadev:
This looks _great_! Thanks so much for putting in this work :+1:.

I noticed a couple things from tinkering with this so far:
- [ ] The mobile collapse feature works great with JS, but becomes unusable without it. I'm not sure how Gatsby handles `noscript`, but I would suggest displaying the menu in the un-collapsed form for mobile users without Javascript.
- [ ] The header tabs don't "stick" to the bottom of the header when the header text collapses ([example](https://imgur.com/eSzMR23))
- [ ] The nav breakpoints probably need to be adjusted: they seem to collapse on top of each other before the "Menu" breakpoint kicks in ([example](https://imgur.com/Reaer6r))

These are all relatively nit-picky compared to all the work you've done here -- thanks so much!

schwartzadev:
I made some updates to implement the header tabs on mobile. (They use horizontal scroll...I think that's what @tealtan had in the mockups. Please correct me if I'm wrong about that!)

image

image

image

kevee:
Thanks for that feedback. I added a style in a noscript to make the mobile menu always appear.

https://5e80a94362237500080f3824--goofy-tesla-43452a.netlify.com/

kevee:
Since this is 90% of the way there, I'm going to merge so we can keep our PR conflicts low. Opening a separate issue to track the wrapping.

PR #230, opened 2020-03-28
Add basic animated/scrubber map

jeremiak:
This isn't totally done, but I'd love a review and to merge if its acceptable. We're missing lots of styling but this is just the bones of the graphic.

Mar-28-2020 15-43-32


goleary:
Just looking at the end result (code comments will be elsewhere)

A couple of issues which you may be aware of:
- distorted play button:
- it should start on the latest day so it contains useful info without having to play
- It would be nice if the animation would stop on the latest day.
- map overflows over content below

I would argue that the current chloropleth map with hoverstate & the info box relays more information to the user than this map.

Take for example Louisiana. It seems as though their problem is no worse than Texas and better than florida when looking at the bubble map:

image

However when you look at the per capita cholorpleth it becomes readily apparent that Louisana is currently much worse off than most states in the country.  This information is somewhat lost in the bubble map.
image

PR #248, opened 2020-03-30
fix(code): Updated redirects

kevee:
Made sure the netlify redirects file still worked.


kevee:
Fixed, api and redirects work:

PR #282, opened 2020-03-31
feat(site): Tiny tweak to use dotenv

friej715:
Very tiny change to make environment variables a bit easier to set with dotenv. (Figured first PR should be small!)


kevee:
Looking good, thanks @friej715 !

PR #286, opened 2020-03-31
Visual tweaks

tealtan:
- Updates the detail text formatting and spacing
- Uses the Container component on the header and footer so that they are the same width as content


kevee:
Looks good, merging.

PR #295, opened 2020-03-31
feat(dashboard-map): multiple colors for chloropleth

goleary:
covid_choloropleth_2
covid_choloropleth

### DIFF
--- a/_src/_assets/js/dashboard-map.js
+++ b/_src/_assets/js/dashboard-map.js
@@ -130,6 +130,7 @@
     const map = svg.append('g')
     const bubbles = svg.append('g')
     const testBubbles = svg.append('g')
+    const legend = d3.select('#map-legend').append('svg')
 
     updateMap()
 
@@ -139,9 +140,7 @@
 
     function updateLegend() {
       if (useChloropleth) {
-        d3.select('#map-legend')
-          .selectAll('*')
-          .remove()
+        legend.selectAll('*').remove()
         return
       }
       const legendData = [
@@ -149,8 +148,6 @@
         parseInt(maxValue * 0.5),
         maxValue,
       ]
-      const legend = d3.select('#map-legend').append('svg')
-
       legend
         .attr('height', 150)
         .attr('width', 150)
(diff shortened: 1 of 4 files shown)


_src/_assets/js/dashboard-map.js
```diff
@@ -103,8 +126,51 @@
...
+      // set up property selector
+      propertyDropdown
+        .selectAll('option')
+        .data(propertyOptions)
+        .enter()
+        .append('option')
+        .text(d => d.name)
+        .attr('value', d => d.value)
+      propertyDropdown.on('change', () => {
+        currentField = propertyDropdown.property('value')
```
jeremiak, line 168: So, this totally works!

But in D3, event handlers work in a sorta idiomatic way. The arguments passed are `(d, i)`. `d` is the data value, so in this case it will be the `object` in `propertyOptions`, so instead of reaching into the DOM for the property value, it could just be `d.value`

_src/_assets/js/dashboard-map.js
```diff
@@ -117,15 +183,17 @@
...
 
     const r = d3
       .scaleSqrt()
       .domain([0, maxValue])
       .range([0, 50])
     const map = svg.append('g')
-    const bubbles = svg.append('g')
-    const deathBubbles = svg.append('g')
+    const testBubbles = svg.append('g').attr('id', 'testBubbles')
+    const bubbles = svg.append('g').attr('id', 'bubbles')
```
jeremiak, line 196: What is the difference between `testBubbles` and `bubbles`?

_src/_assets/js/dashboard-map.js
```diff
@@ -134,68 +202,88 @@
...
+            width: 300,
+            marginTop: 8,
+            tickFormat: '~s',
+            spaceBetween: 2,
+            tickSize: 0,
+          }),
+        )
+      } else {
+        //bubble legend
+        const legend = d3
```
jeremiak, line 227: Looking forward to getting this refactored in the `d3Legend` so we can use it in two places 🎉 🎉 🎉

_src/_assets/js/dashboard-map.js
goleary: the map now has concentric circles, one for test #s and one for positives.  I was having trouble using a single element to contain them so i just ended up using two.

_src/_assets/js/dashboard-map.js
goleary: good to know, I think I was following your method for the start-stop checkbox.

_src/_assets/js/dashboard-map.js
goleary: agreed

PR #301, opened 2020-04-01
Fix(design): fix table border size mismatch.

ryansloan:
Fixes https://github.com/COVID19Tracking/website/issues/109 ("Double border on tables"). Updated to apply borders to `.state-flex-container` rather than adding up individual borders on cells.

**Before (zoomed in)**
Before - zoomed

**After (zoomed in)**
After - zoomed

**Before (full size)**
Before - full size

**After (full size)**
After - full size

**Before (mobile size)**
Before - mobile view

**After (mobile size)**
After - mobile view

### DIFF
--- a/_src/_assets/css/_tables.scss
+++ b/_src/_assets/css/_tables.scss
@@ -41,7 +41,7 @@ tbody tr:nth-child(even) {
   display: flex;
   flex-wrap: wrap;
   flex-direction: column;
-
+  border: 1px solid #ddd;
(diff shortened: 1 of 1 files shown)


kevee:
Hi, @ryansloan  - sorry but we're about to merge our `gatsbyjs` branch into master, which will totally break this PR. Do you want to review the table design and submit a PR if the table designs are still not up to snuff?

Thanks!

PR #304, opened 2020-04-01
fix(dashboard): updated styles for button and fixed overlapping grid …

melbee:
- Updated collapse button styles to match Julia's sketch design for death bar chart
- Fixed grids lines that were previously appearing over the bars for death bar chart
- Removed ticks from horizontal death bar chart
- Updated color for both vertical and horizontal death bar charts
- Updated label format for area chart 
Screen Shot 2020-04-01 at 7 33 52 PM

Screen Shot 2020-04-01 at 7 34 02 PM

### DIFF
--- a/_src/_assets/js/d3-area-chart.js
+++ b/_src/_assets/js/d3-area-chart.js
@@ -103,6 +103,13 @@ function d3AreaChart({
     .axisLeft()
     .scale(yScale)
     .ticks(4)
+    .tickFormat(function(d){
+      if(d/1000000 < 1){
+        return d/1000000;
+      } else {
+        return d/1000000 + " Million";
+      }
+    })
   const xAxisG = axi
     .append('g')
     .classed('axis x-axis', true)
--- a/_src/_assets/js/dashboard-charts.js
+++ b/_src/_assets/js/dashboard-charts.js
@@ -309,28 +309,15 @@ d => d.negativeIncrease + d.positiveIncrease,
     const chartContainer = d3.select(
       '#chart-states-current-death-total',
     )
-    const hed = chartContainer
-      .append('h3')
-      .classed('chart-hed', true)
 
-    const chart = chartContainer
-      .append('div')
-      .classed('chart', true)
-      .classed('no-y-axis-domain', true)
-
-    const expand = chartContainer
-      .append('div')
-      .classed('chart-expand-note', true)
-    const source = chartContainer
-      .append('div')
-      .classed('chart-api-note', true)
+    const chart = chartContainer.select('.chart')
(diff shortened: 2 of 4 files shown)


_src/dashboard.md
```diff
@@ -82,7 +82,18 @@ nav: Dashboard
     <div class="graphic-text">
        <p>Though this is a national crisis, each state is reporting data differently. We are tracking numbers from each state, though the quality and frequency of reports varies significantly.</p>
     </div>
-    <div class="graphic" id="chart-states-current-death-total"></div>
+    <div class="graphic" id="chart-states-current-death-total">
+      <h3 class="chart-hed">Total deaths by State</h3>
+      <div class="chart no-y-axis-domain"></div>
+      <div class="graphic-footer">
+        <div class="chart-expand-button"></div>
```
jeremiak, line 89: If this is serving as a button, that is to say that a viewer can click it and the page will respond, we should make it an actual button so that it can be interacted with from a keyboard/assistive technology. Not worth preventing a merge, but definitely something we want to clean up soon in a subsequent PR.

_src/_assets/js/dashboard-charts.js
```diff
@@ -379,21 +364,26 @@ d => d.negativeIncrease + d.positiveIncrease,
...
+      const tickSelector = id + ' .y-axis-group .tick'
+      const chart = container.select('.chart-group')
+      
         d3.selectAll(tickSelector).each(function(d) {
           const tick = d3.select(this)
           const line = tick.select('line')
-
-          line.attr('x1', container.node().clientWidth * 0.78)
+          
+          // grid lines for vertical bar charts only
```
jeremiak, line 374: 👍

_src/_assets/js/dashboard-charts.js
```diff
@@ -13,6 +13,7 @@
 
   const totalColor = '#585BC1'
   const positiveColor = '#FFA270'
+  const deathsBarColor = '#3F4856'
```
jeremiak, line 16: Maybe this should just be `deathsColor` to match the other variable naming scheme?

_src/_assets/js/d3-area-chart.js
```diff
@@ -103,6 +103,13 @@ function d3AreaChart({
     .axisLeft()
     .scale(yScale)
     .ticks(4)
+    .tickFormat(function(d){
+      if(d/1000000 < 1){
+        return d/1000000;
+      } else {
+        return d/1000000 + " Million";
```
jeremiak, line 110: This is a fine, approach but I actually think we just want this to happen on the first tick and not any tick over a million. Also I think the "million" should be lowercase.

Issue #313, opened 2020-04-01
Slightly weird alignment on homepage


friej715:
Noticed on my iPhone 6S, Safari. Visible to varying degrees on different Chrome window sizes.

![IMG_1641](https://user-images.githubusercontent.com/1028403/78161210-1d1f1b80-7413-11ea-97cb-b10e3412e14f.jpg)

beep:
It looks like each item is a flexbox’d div, with an inner div styling the bullet. Switching over to lists and background images might simplify things a bit. (Or, if the `ul` switch isn’t an option, background images.)

beep:
(All that said, the _super_ quick fix would be something like `flex: 0 0 30px` on the icons)

kevee:
Fixed!

PR #340, opened 2020-04-02
fix(site): Tweaking homepage's Get Involved

friej715:
Another teeny atomic PR, this time re: #313. Also took out what appears to be a typo'd/extraneous class.


kevee:
Awesome, @friej715, that has been really irking me for some reason. Thanks!

Issue #311, opened 2020-04-01
Italicize press pub titles


kissane:
Can we (post-launch) get cite tags or some such around the publication titles in our press links, and make sure they're styled in italics? News orgs have such strong feelings about their italics.

kevee:
As an ardent believer in Chicago style, I can't agree more.

schwartzadev:
Closing this -- merged #344

PR #344, opened 2020-04-02
Cite tags and italicizing for press

friej715:
For #311. Let me know if this separate scss file is overkill for the job--this formatting has to apply to both homepage and press page, so I figured it might be handy to just have one place.


schwartzadev:
Looking great!
I'm 100% on board with breaking up our scss 😄 
image
Merging...

PR #347, opened 2020-04-02
Simplify Gatsby setup in README

olivierlacan:
Let's not list environment variables in the README when we can put that all in
a single place and let's not encourage people to use bash profile for environment
variable management when we have dotenv in the dependencies taking care of that
for us.

I've simplified the installation instructions accordingly but eventually I
think the `cp` operation could be covered by a setup script. I'm just not
super familar with external npm setup scripts, I'd want it to be a bash
script that outputs some feedback, not something that just copies the
`.env.example` file and leaves it in a broken state without any feedback.


kevee:
These are great additions - once we get the API moved completely into the website I think we should create a "setup" script in pacakage.json that pulls down sample API data, prompts for env vars, etc. We'll get there!

PR #375, opened 2020-04-02
feat(docs): small presslist update

friej715:
Another atomic lil' update for Storybook docs. Next PR will have more :)


kevee:
Looking good, thanks!

Issue #266, opened 2020-03-31
API: Migrate API as a build step before site compiles


webmasterkai:
I think the first Netlify function is probably a web-hook endpoint that fetches the current data for us and states from the sheets and saves it to the repo after some slight processing.

After that we'll add a build step to the website that takes that file and splits it out into a bunch of other files to serve as the basis of the API.

Possible structure:

- /data/current/us.json (replaces /api/us)
- /data/current/us.csv (replaces /api/us.csv)
- /data/current/states.json (replaces /api/states)
- /data/current/states.csv (replaces /api/states.csv)

Other pages that were query args could be like:

- /data/current/state/AK.json (replaces /api/states?state=AK)

We will then update the old Cloudlfare endpoints to act as a simple 301 redirect to the new static file locations based on the request.

webmasterkai:
We are going to move forward with building all json asset/data files at site compile time. The `/functions` dir will have code related to any interactive API code like GraphQL. The `/build` dir will contain code related to building the json files. Both will have dedicated package.json files to maintain dependencies.

I'm now thinking a structure more like:
- /data/us/current.json
- /data/us/daily.json
- /data/states/current.json
- /data/states/daily.json
- /data/state/[state]/current.json
- /data/state/[state]/20200329.json

pspeter3:
Perhaps we want to rename `/build` to `/generate` since `/build` is sometimes used as an output directory?

The structure looks good though wondering if we want `/data/state/:state/:yyyy-mm-dd.json`?

Also do we want `/data/us/:yyy-mm-dd.json`?

pspeter3:
I assume there will also be a CSV equivalent for each of these?

pspeter3:
I would like to propose `/data/:version` as a prefix. That way we can write multiple versions. Unsure how we do this for GraphQL. I think GraphQL has its own scheme for deprecations.

pspeter3:
Based on discussions in Slack, the team thinks that not versioning the API and only appending fields is the easier way to go. This means that the generated data and GraphQL api will follow the same versioning and deprecation scheme.

webmasterkai:
Okay, I'm tired but I think I got most of the code that creates API assets added to the repo https://github.com/COVID19Tracking/website/tree/api-2/build

kevee:
Initial feels is: run npm run [get-api-thing] - then it writes to like _data , then Gatsby builds off of that data, then we just copy _data into /public/api

kevee:
[...]

PR #382, opened 2020-04-03
test(api): Add jest tests for API quality

### LINKED ISSUE #381: Add some jest tests that pull all the main API endpoints and check data quality.
We also need to add some jest tests or something that pulls all the main API endpoints and does some data quality heuristics (number of items in array? object composition?). This would run after the build and before Gatsby runs, so that if the Jest test fails, the build is flagged as failing and we don't break anything.

_Originally posted by @kevee in https://github.com/COVID19Tracking/website/issues/266#issuecomment-607249211_

gitjeff05:
Fixes #381 

This is a first stab at adding some testing around [API quality as referenced here](https://github.com/COVID19Tracking/website/issues/266#issuecomment-607249211). These first four tests add some basic data quality checks ensuring:

1.) Length of items in daily (us & states) array.
2.) Latest daily state data has at least 50 items.
3.) Latest daily states is subset of states (`./build/datasources/stateNames.js`)

While these are by no means comprehensive, it is a step forward. There are a couple of caveats:

1.) Since pulling the data from the spreadsheet during build time requires Google API keys, only privileged users with credentials will have the ability to reliably run these tests. Others, will have to manually pull from the site to the `./public/api` directory. I have created a bash script for that not included in this PR. We could also just mock it so that at least it would not fail for other users.
2.) Static thresholds will become stale. This could possible be overcome with utilizing Jest snapshots or some other method.

### DIFF
--- a/build/datasources/stateNames.js
+++ b/build/datasources/stateNames.js
@@ -73,4 +73,5 @@ module.exports = {
   nameByCode: _.propertyOf(states),
   fipsByCode: _.propertyOf(fips),
   codeByName: _.propertyOf({ ...stateNames, ..._.invert(states) }),
+  allStates: states,
 }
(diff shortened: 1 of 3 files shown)


gitjeff05:
@kevee fixed API path to `public/api/v1`. Should be good to go now.

PR #388, opened 2020-04-03
fix(prettier-config): enable building on windows

goleary:
without this fix: ` error  Delete '␍'  prettier/prettier` for every newline of every file on windows:
image
https://stackoverflow.com/questions/53516594/why-do-i-keep-getting-delete-cr-prettier-prettier

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -3,4 +3,5 @@ module.exports = {
   singleQuote: true,
   trailingComma: 'all',
   semi: false,
+  endOfLine: 'auto',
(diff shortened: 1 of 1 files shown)


kevee:
Would these docs be helpful to link to?

https://www.gatsbyjs.org/docs/gatsby-on-windows/

goleary:
it was a git issue rather than gatsby. I think you're good to leave everything as is and if anyone else complains about the same issue we can look at a doc or setting update.

PR #389, opened 2020-04-03
API redirects & CORS

### LINKED ISSUE #384: API: Add some CORS headers to netlify config for json files
I think something like the following added to netlify.toml

```
[[headers]]
  for = "/*.json"
  [headers.values]
    Access-Control-Allow-Origin = "*"
```

### LINKED ISSUE #386: API: Add Redirects to netlify config
Cloudflare could handle query args so you could get just the info you wanted like https://covidtracking.com/api/us/daily?date=20200306

Now we just generate the json file for every date https://covidtracking.com/api/v1/us/20200306.json

Netlify allows for some redirects that are able understand the search args so I think there’s some redirect configs that will let us forward correctly

kevee:
Added API redirects to the new "V1" prefixed API.
Added CORS header

Fixes #386 
Fixes #384

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,8 @@
 [build]
   publish = "public"
   command = "npm run build"
+
+[[headers]]
+  for = "/api/*.json"
+  [headers.values]
+    Access-Control-Allow-Origin = "*"
--- a/static/_redirects
+++ b/static/_redirects
@@ -1,6 +1,6 @@
 /api/states state=:state                      /api/v1/states/:state/current.json
-/api/states/daily state=:state                /api/v1/states/:state/daily.json
 /api/states/daily state=:state date=:date     /api/v1/states/:state/:date.json
+/api/states/daily state=:state                /api/v1/states/:state/daily.json
 /api/states/info state=:state                 /api/v1/states/:state/info.json
 /api/us/daily date=:date                      /api/v1/us/:date.json
 /api/states state:state                       /api/v1/states/:state/current.json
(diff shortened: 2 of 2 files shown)


kevee:
I tested a few original URLs off this deploy preview and I think we're looking good. @webmasterkai  - wanna check?

https://deploy-preview-389--upbeat-lovelace-3e9fff.netlify.com/

kevee:
So we still want `/api/states` to get proxied to the API endpoint?

webmasterkai:
If we set a redirect as `/api/states` it doesn't handle query args that aren't anticipated like `?grade=B`. Checking the logs a bunch of endpoints have various args that Cloudflare still needs to resolve.

kevee:
Got it, thanks!

kevee:
If there is a list of used query parameters, I'm fine with putting those in redirects in the interest in getting the bulk of traffic off to static files.

Any "new" parameters folks use, of course, won't work, but that means they read the docs and see that things have changed.

```

/api/states/daily state=:state date=:date     /api/v1/states/:state/:date.json
/api/states/daily state=:state                      /api/v1/states/:state/daily.json
/api/states/daily state=:state grade:grade                      https://covid.cape.io/states?state=:state&grade=:grade 200
... ad nauseum with all possible query param combos
/api/states/daily                /api/vi/states/daily.json
 

```

kevee:
If nothing else, what we can do here is use Netlify functions to just take those requests, read the static JSON, filter, and respond with the data.

Issue #325, opened 2020-04-01
People are missing the links to states on the big /data page


kissane:
I feel like we have a dropdown comped in at some point? We're getting sad tweets about missing these links from mobile users.

schwartzadev:
Which links? 🤔 Here's what I'm seeing on mobile:

![image](https://user-images.githubusercontent.com/18607205/78194178-ab0dfd00-7439-11ea-83ea-4ac9ad6948c0.png)

kissane:
Sorry, the quick links so people don't have to scroll forever. Someone objected to them and we nixed them, but I'm getting a lot of tweets and DMs. We had a dropdown suggested at some point to replace them?

schwartzadev:
Ah yes -- the navigation. Silly me. :man_facepalming:  I'll throw something together for now, but @beep @tealtan @jasonsantamaria maybe you could design something beautiful for this down the line?

kevee:
There's a dropdown in the sketch file that looks like a regular select element - I'd like to use something like [Reach's combobox](https://reacttraining.com/reach-ui/combobox) because it's got great a11y features

schwartzadev:
Oooh Kevin that looks great -- are you thinking we use combobox to:then jump to the selected state (i.e. a user inputs the state name/abbreviation and then clicks on the result) or that we would filter based on the combobox result (i.e. a user inputs "New" and all results disappear except for New York, New Jersey, etc.)?

In the meantime, I just pushed #342  -- Kevin, if you're on board, please merge it for the interim 👍

kevee:
I think option 1 - I'm merging your PR in a bit.

beep:
> There's a dropdown in the sketch file that looks like a regular select element - I'd like to use something like Reach's combobox because it's got great a11y features

Sounds good. I’m not familiar with combobox; what’s the experience like if JS fails to load?

schwartzadev:
...or if you don't load JS intentionally 😉

beep:
(I knew I liked you, @schwartzadev 😉)

kevee:
We can use the current select list that @schwartzadev built, hide it with a style in a `<noscript>` tag - the combobox component could degrade to a select element, but that would be useless without JS anyhow because you can't monitor events.

beep:
> hide it with a style in a <noscript> tag

I think that’ll work if JS is actively disabled, but not if there’s some other condition that prevents JS from executing, right?

[...]

PR #391, opened 2020-04-03
feat(design): Added jump-to-state combobox

kevee:
Related to #325 

This adds a Combobox element to the right/bottom of the data page that jumps people to a selected state.

The existing state nav element is retained and hidden except with JS turned off.

### DIFF
--- a/src/pages/data/index.js
+++ b/src/pages/data/index.js
@@ -1,3 +1,5 @@
+/* eslint jsx-a11y/label-has-associated-control: 0 */
+
 import React, { useState } from 'react'
 import { graphql } from 'gatsby'
 import {
@@ -73,11 +75,7 @@ const StatesNav = ({ stateList }) => {
     }
     const results = []
     stateList.forEach(({ node }) => {
-      if (
-        `${node.state} - ${node.name}`
-          .toLowerCase()
-          .search(term.toLowerCase()) > -1
-      ) {
+      if (node.name.toLowerCase().search(term.toLowerCase()) > -1) {
         results.push(node)
       }
     })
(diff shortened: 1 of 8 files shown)


beep:
A few things I noticed!

- I’m not quite clear on how the sorting works? The first result was often unrelated to the letter I typed in. (Here’s [a screenshot for `m`](https://user-images.githubusercontent.com/124493/78376254-0bb54b00-759c-11ea-92e0-16c59a8b69c4.png) and [a screenshot for `c`](https://user-images.githubusercontent.com/124493/78376242-0821c400-759c-11ea-8ecd-14304bdd49d6.png).) (If the two-letter abbreviations are the culprit, I think we could probably drop them.)
- Can “Jump to state” be a `label` associated with the rendered field?
- Three usability questions!
  - I wonder if the element could better describe the action required? Like, could/should the label read “Type a state’s name to jump to it”?
  - Is there any way to get the full list to appear when the text field first gains focus?
  - If not, it’d be really great if pressing the down arrow revealed the full set of options. (Sort of how [the native `datalist`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) element works.)
- Smol feature request: could we make the module containing “Totals by state” and the jump menu sticky?

kevee:
OK, I:

Removed the state code, that was causing errors
Made the box always appear and boldify the found text
Changed to a label element for the input
Changed the label text

kevee:
...and using a sticky head covers the title of the state when you jump to its anchor...  🤔

kevee:
Yes it would!

src/pages/data/index.js
```diff
@@ -34,19 +44,110 @@ const StateList = ({ states, stateData }) => {
...
+            return node.name === selectedItem
+          })
+          if (stateId && typeof window !== 'undefined') {
+            window.location.hash = `state-${stateId.node.state.toLowerCase()}`
+          }
+        }}
+      >
+        <ComboboxInput
+          id="jump-to-state"
+          placeholder="State or territory"
```
schwartzadev, line 112: :+1: for not using "state" :smile:

kevee: Having lived in a territory, I try not to be "state normative."

kevee:
OK, updates:

- Switched to searching as an autocomplete (start of state name) rather than any match (i.e. "al" matches "C**al**ifornia)
- Disabled browser autocomplete
- `scroll-margin` was not working when updating the window location with Javascript. Used an older trick to not cover state headers when navigating.

Note that we don't have a sticky header in mobile because it was taking up valuable real-estate. I defer to other's judgement on that.

schwartzadev:
@kevee that was the issue with JS -- thanks 👍 

Yay! The autocomplete algorithm looks good -- this is a great feature to add!

My only suggestion at this point is pretty minor: when there is only one result remaining, I expect that pressing `enter` would jump to it. (Similar to navigating HTML's `select` with a keyboard.) However, that's not the current behavior; right now, pressing enter does nothing. (i.e. try entering `ca`. The only result is California, yet pressing enter does not jump to California.)

This is _super_ marginal -- I'm on board if you'd like to merge and move this to an issue instead of waiting to merge.

kevee:
@schwartzadev - Thanks for the feedback! I just added your suggestion about Enter key when there is one result also jumping you to that result.

schwartzadev:
Yes! This looks phenomenal 😄 The only thing that I found as an issue was the nav bar padding seems to break off on the right side between 1080px and 640px:

image

I'll see if I can fix it 👍

schwartzadev:
Hmm it looks like it's because of the container width but the tables scroll at that width...
image

schwartzadev:
b79cce1145bfde2affb4500ab8748f0865e4bfa4 helps this but extends the bottom border to the right edge. I'll look into a fix.
image

schwartzadev:
...right when I thought I fixed this...the table borders are peeking out:
image

kevee:
Thanks for all the helpful tweaks, @schwartzadev !

PR #396, opened 2020-04-03
Convert area chart to React

jeremiak:
Bring over our area chart code from the previous version so that its a React component used on the dashboard page.

Screen Shot 2020-04-03 at 4 27 26 PM


src/pages/dashboard/index.js
```diff
@@ -1,12 +1,77 @@
...
+        },
+      ]
+    })
+    .flat()
+}
+
+const DashboardPage = () => {
+  const [data, updateData] = useState([])
+
+  useEffect(() => {
```
goleary, line 38: it will get complicated if we're fetching all of the data from this page but the graphql queries will simplify that a ton (each component will specify which data it needs).

src/pages/dashboard/index.js
```diff
@@ -1,12 +1,77 @@
...
+    async function fetch() {
```
goleary, line 39: probably best not to name a function after a existing brower api. perhaps `fetchData` instead

src/pages/dashboard/index.js
```diff
@@ -1,12 +1,77 @@
...
+          label: 'Positive',
+          value: d.positive,
+        },
+      ]
+    })
+    .flat()
+}
+
+const DashboardPage = () => {
+  const [data, updateData] = useState([])
```
goleary, line 36: the idiomatic way is `const [<value>, set<Value>] = useState([])` instead of `update`

src/pages/dashboard/index.js
```diff
@@ -1,12 +1,77 @@
...
-const DashboardPage = () => (
-  <Layout title="Visual Dashboard">
-    <p>Test</p>
-    <Map />
-  </Layout>
-)
+import AreaChart from './_AreaChart'
+// import Map from './_Map'
+
+const parseDate = timeParse('%Y%m%d')
```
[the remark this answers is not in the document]
melbee: + 1

src/pages/dashboard/index.js
jeremiak: Great catch

jeremiak:
Start adding back small multiples, though I'm not totally sure I'm organizing this CSS properly
Screen Shot 2020-04-07 at 12 51 07 PM

src/pages/dashboard/_util.js
```diff
@@ -0,0 +1,75 @@
...
+import { timeFormat, timeParse } from 'd3-time-format'
+
+export const formatDate = timeFormat('%b. %e')
+export const formatNumber = format(',')
+export const parseDate = timeParse('%Y%m%d')
+export const calculateTotal = d => {
+  return d.positive + (d.negative || 0)
+}
+export const getStateName = abbr => {
+  const names = {
```
goleary, line 12: The two letter state abbreviation to state name is contained within the data returned by this endpoint: https://covidtracking.com/api/v1/states/info.json

I'd say it's better to use that than have a hrardcoded mapping.  Perhaps this is something we can remove after converting the code to use the gatsby data.

src/pages/dashboard/index.js
goleary: @melbee's suggestion was that we don't do any data fetching inside `index.js` and instead have a wrapper component for each chart type that fetches the data, does any presentation logic & then passes the data down to the relevant chart components.

src/pages/dashboard/_util.js
jeremiak: I'm down! Though I'd love to figure out how to do the GraphQL query so it happens at build time and doesn't require another request from the browser

Issue #328, opened 2020-04-01
change colors of map button 


julialedur:
let's make the selected button with white background and black text and stroke. it's cleaner this way 

<img width="392" alt="Screen Shot 2020-04-01 at 3 55 13 PM" src="https://user-images.githubusercontent.com/39416238/78180738-320aa780-7431-11ea-9575-58c1810d5713.png">

kevee:
@goleary  & @julialedur  - Quick heads up on colors:

We actually have [official color swatches](https://covid19tracking.github.io/website-docs/?path=/story/colors--main-colors).

Those were unfortunately not selected with the idea of graphs/maps etc. in mind. I'd suggest that you reach out to @jasonsantamaria. Come up with a color swatch and we can add them in!

And then to use them in your code, you can do this:

In Javascript (like graphs and such):
```javascript
import colors from '../scss/colors.scss'

const amazingChartBackground = colors.secondaryDark
```

In SCSS:
```scss
@import '../colors.scss'

.awesome-chart {
  background: $color-secondary-dark;
}
```

julialedur:
Hey Kevin! 
Thanks for the heads up! 
Will talk to Jason about it :)

goleary:
the original issue is resolved. @julialedur perhaps you should open a new one to start using the swatches once #440 is done.

PR #402, opened 2020-04-04
feat(dashboard-gatsby):add bubble map with tooltips

goleary:
Choropleth code also present but I haven't added any of the toggles/the slider back yet.
that's next.


src/pages/dashboard/.Map.js
```diff
@@ -0,0 +1,155 @@
+import React, { useState, useMemo, useEffect } from 'react'
+
+import { json } from 'd3-fetch'
+import { nest } from 'd3-collection'
+import { sum } from 'd3-array'
+
+import _Map from './charts/_Map'
+import { path } from './charts/_Map'
```
jeremiak, line 8: Nitpick, but I've usually seen this done as `import _Map, { path } from './charts/_Map'` - there an advantage either way?

src/pages/dashboard/.Map.js
```diff
@@ -0,0 +1,155 @@
...
+
+  const sumChoro = useMemo(
+    () => joinedData && sum(joinedData.features, d => getValue(d)),
+    [joinedData, getValue],
+  )
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const stateData = await json('https://covidtracking.com/api/states/daily')
+      setRawStateData(stateData)
```
jeremiak, line 78: So its cool to update two different states in the same function without trigger a double render pass?

src/pages/dashboard/.Map.js
```diff
@@ -0,0 +1,155 @@
...
+
+import { json } from 'd3-fetch'
+import { nest } from 'd3-collection'
+import { sum } from 'd3-array'
+
+import _Map from './charts/_Map'
+import { path } from './charts/_Map'
+import StatesWithPopulation from './data/_state-populations'
+
+import { formatDate, formatNumber } from './.utils'
```
jeremiak, line 11: What does the leading period signify in the filename here?

goleary: i was unaware of this but it looks like either a `.` or `_` stop gatsby from trying to create a page.  Otherwise if you have `dashboard/utils.js` it tries to create a route at `/dashboard/utils/`

I was following the organizational suggestions set forth by @melbee altho I'm not sure i did it exactly as she suggested.  I'm going to have her review this as well to make sure we're on the same page.

src/pages/dashboard/.Map.js
goleary: no, i actually prefer your way, just forgot about it this time

src/pages/dashboard/.Map.js
melbee: If I included a period before the file name that’s a typo on my end! 

import { formatDate, formatNumber } from './utils' 

Above is what I intended

src/pages/dashboard/charts/_Map.js
```diff
@@ -0,0 +1,297 @@
...
+  death: scaleThreshold(colorLimits.death, schemeGreys[8]),
+  positive: scaleThreshold(colorLimits.positive, schemeOranges[8]),
+  totalTestResults: scaleThreshold(
+    colorLimits.totalTestResults,
+    schemePurples[8],
+  ),
+}
+
+// should be imported from constants file
+const colors = {
```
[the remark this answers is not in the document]
goleary: I usually do the same, but it sounds like the rest of the site is using colors as described here: https://github.com/COVID19Tracking/website/issues/328#issuecomment-610121230

I'm going to leave this for now until we get our colors into this which I think @julialedur is working on.

src/pages/dashboard/.Map.js
goleary: that was my presumption, and I just confirmed it to be sure.

`useEffect` runs between renders so calling `setSate` multiple times within it will only trigger a single render after the effect is done.

you can see [here](https://codesandbox.io/s/tender-dijkstra-sw6c9?file=/src/index.js). (one render from before the effect is run and one render after the effect is run no matter how many states I set in the effect).

PR #412, opened 2020-04-04
Fix active tab state when URL has trailing slash

vijaysharwar:
When navigating using the site tabs/UI, the URLs do not have trailing slashes, which ensures the Gatsby `Link` components render with correct aria-current attributes and selected states. The current page's tab is highlighted.

However, if you directly navigate with a trailing slash, such as:

[https://covidtracking.com/data/](https://covidtracking.com/data/)

The `Link` matching logic fails. Adding the [gatsby-plugin-remove-trailing-slashes](https://www.gatsbyjs.org/packages/gatsby-plugin-remove-trailing-slashes/) plugin fixes the issue, but only works in production mode.

To test locally, run `gatsby build` and serve the /public directory using local static file server.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -33,6 +33,7 @@ module.exports = {
     `gatsby-plugin-sass`,
     `gatsby-transformer-yaml`,
     'gatsby-plugin-eslint',
+    `gatsby-plugin-remove-trailing-slashes`,
     {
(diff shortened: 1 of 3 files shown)


kevee:
Thanks, @vijaysharwar  - I'm going to do more testing with Netlify to make sure we don't have redirect loops, then we should be good to merge.

PR #417, opened 2020-04-05
fix(design): Changed state display to order states alphabetically.

### LINKED ISSUE #416: Request: Sort data page by name, not abbreviation
I realize the states are sorted by their abbreviations and that makes sense for the navigation table up top, but when scrolling, the states should be in order alphabetically.

jgulian:
This pull request fixes #416, by sorting state navigation by the abbreviations, and sorting the state data by the state name, in /data/.

### DIFF
--- a/src/pages/data/index.js
+++ b/src/pages/data/index.js
@@ -83,7 +83,7 @@ export const query = graphql`
         }
       }
     }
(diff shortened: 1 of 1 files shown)


jgulian:
This is a work in progress

PR #423, opened 2020-04-06
Remove US total outcomes data

schwartzadev:
Alexis put this out to the website team earlier today:
> Because some states report current and some states report cumulative hospitalization/ICU numbers, we a fundamental mismatch in what these data represent. We can't just push currents across into cumulatives to get a cumulative number [...] The only real answer, I think, is to not provide a national summary of outcomes data.

This PR **does** remove the outcomes data from the US summary table
This PR **does not** remove the outcomes data from individual states (it is retained on both the data index, under each state and on the individual state pages)

[Relevant Slack thread](https://covid-tracking.slack.com/archives/CUYAS1M25/p1586199198317900)

We'll need to update the contentful snippet (`dataSummaryFootnote`) for this update as well
----

### DIFF
--- a/src/components/common/summary-table.js
+++ b/src/components/common/summary-table.js
@@ -3,7 +3,7 @@ import Table from './table'
 import '../../scss/components/common/summary-table.scss'
 import thousands from '../../utilities/format-thousands'
 
-export default ({ data, lastUpdated }) => (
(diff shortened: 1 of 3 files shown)


webmasterkai:
I like and appreciate the `showOutcomes` flag.

PR #427, opened 2020-04-07
Refactor SCSS to use modules

### LINKED ISSUE #390: Refactor SCSS files
In our great haste, we have a few SCSS files that are spreading global classes around. Let's refactor these to:

- Wrap pages/components in more classes - We can just apply more classes and always have each .scss file scoped to that class. We're most of the way this direction
- Use [SCSS modules](https://www.gatsbyjs.org/docs/css-modules/) - **Pro:** This will *force* us to use modules. **Con:** Harder for folks to just jump in
- **Styled components in Emotion** - **Pro:** Um, I like it? **Con:** Hard for new devs to jump in

kevee:
Fixes #390 #453

Refactored all the SCSS, moved them into modules, and removed a few unused files.


src/scss/normalize.css
```diff
@@ -31,30 +31,6 @@ main {
...
- * 1. Add the correct box sizing in Firefox.
- * 2. Show the overflow in Edge and IE.
- */
-
-hr {
-  box-sizing: content-box; /* 1 */
-  height: 0; /* 1 */
-  overflow: visible; /* 2 */
-}
-
```
beep, line 57: Just curious: what was the reason for this change? (Thinking ahead to any future updates of normalize we might need to pull in.)

src/scss/normalize.css
```diff
@@ -163,7 +139,6 @@ select,
 textarea {
   font-family: inherit; /* 1 */
   font-size: 100%; /* 1 */
-  line-height: 1.15; /* 1 */
```
beep, line 166: Same question here as on https://github.com/COVID19Tracking/website/pull/427/files#r405601317, I guess! Was this line-height creating some issue we couldn’t override?

CONTRIBUTING.md
```diff
@@ -12,6 +12,12 @@ The project README file includes all the information you need to build a local v
 
 We use the [Airbnb style guide](https://github.com/airbnb/javascript), except with semicolons turned off. We suggest using Prettier in your environment to make sure your code files are all balanced and beautiful.
 
+### Style and SCSS Modules
+
+We use Gatsby's [CSS modules](https://www.gatsbyjs.org/docs/css-modules/), but with Sass files. Every Sass file in `/src/scss` should be in a file that mirrors the structure of their parent component. For example, a component at `/src/components/beetle.js` that needs styling should have a corresponding SCSS file at `/src/scss/components/beetle.module.scss`.
```
beep, line 17: Thinking through this a bit: maybe it’d make sense to consolidate styles alongside their components? So rather than

- `/src/components/beetle.js`
- `/src/scss/components/beetle.module.scss`

maybe something like


- `/src/components/beetle/index.js`
- `/src/components/beetle/style.module.scss`

?

(Don’t feel strongly about this, and this is not a blocker for merge; just curious!)

src/components/layout/header.js
```diff
@@ -65,25 +65,31 @@ const Header = ({ title, noMargin, hasHero, navigation }) => {
     <>
       <DevelopmentWarning />
       <header
-        className={`site-header ${showMobileMenu ? 'show-mobile-menu' : ''} ${
-          noMargin ? 'no-margin' : ''
-        }`}
+        className={`${headerStyle.siteHeader} ${
```
beep, line 68: Do we need a human-readable `class` here?

src/components/layout/header.js
```diff
@@ -3,12 +3,12 @@ import { Link, useStaticQuery, graphql } from 'gatsby'
...
 import DevelopmentWarning from './development-warning'
-import Hero from '../common/hero'
+import Hero from './hero'
 import ProjectLogo from '../../images/project-logo.svg'
-import '../../scss/components/header.scss'
+import headerStyle from '../../scss/components/header.module.scss'
 
 const HeaderTabs = ({ navigation }) => (
-  <div className="header-tabs">
+  <div className={headerStyle.headerTabs}>
```
beep, line 11: Do we need a human-readable `class` here?

src/components/common/container.js
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
-import '../../scss/components/common/container.scss'
+import containerStyles from '../../scss/components/common/container.module.scss'
```
beep, line 2: When importing style modules, is there a benefit to assigning a component-specific name to the import? That is, using `containerStyles` or `detailTextStyles` instead of something generic like `styles`?

(I haven’t gatsby’d before; I’m just asking for my edification here!)

src/scss/normalize.css
kevee: h1 was because header height is actually controlled by Typography right now... hr should go back. I'll fix that.

src/components/common/container.js
kevee: I think I just always err on long variable names, plus some pages *could* have multiple sheets in the future.

CONTRIBUTING.md
kevee: I'm already ripping the site's SCSS files apart, so I'll change them all to adjacent files very soon.

kevee:
No worries, I'm going to clean it up and merge with #447 tomorrow am.

PR #428, opened 2020-04-07
Convert dashboard viz to use Gatsby

jeremiak:
We're making a dashboard page with a bunch of charts and maps to show off our data and how to use it! This is a work in progress, but we'd love any feedback you have!

You can see the latest deploy of this branch at [this URL](https://deploy-preview-428--upbeat-lovelace-3e9fff.netlify.com/dashboard).


src/pages/dashboard/index.js
```diff
@@ -0,0 +1,95 @@
...
+    .flat()
+}
+
+const DashboardPage = () => {
+  const [usDaily, setUsDaily] = useState([])
+  const [stateDaily, setStateDaily] = useState([])
+
+  useEffect(() => {
+    async function fetchData() {
+      const usDailyReq = await json('https://covidtracking.com/api/us/daily')
```
kevee, line 55: Since Gatsby already has all the Daily, etc. data in it's database, you don't even need to do fetch request to the API - you can just query it directly.

src/pages/dashboard/map-container.scss
```diff
@@ -0,0 +1,144 @@
+//this should be hoisted to a dashboard constant .scss file
+$positive-color: #e5a968;
+$total-color: #696dc2;
+$death-color: #404856;
+
+#state-map {
```
kevee, line 6: Use classes only - we're moving to CSS modules.

src/pages/dashboard/map-container.scss
```diff
@@ -0,0 +1,144 @@
+//this should be hoisted to a dashboard constant .scss file
+$positive-color: #e5a968;
+$total-color: #696dc2;
+$death-color: #404856;
```
kevee, line 4: I think all these colors should become variables in the `colors.scss` file.

kevee:
There was a change to the spreadsheet that broke builds, so you might want to merge in master to get your previews working in this PR.

gitjeff05:
Good job, this is awesome! A couple of thoughts:

1.) Some charts look crammed on mobile and desktop view. They could fill the page a bit more. See attachment.

2.) From a content perspective, dashboard and "tips for data viz" are two different entities. Intermixing them sends strange signals (_"Is this a blog post on data viz best practices or a dashboard?"_). Why not let the dashboard be primary with best practices sprinkled on in context? For example, an info icon next to the choropleth on hover reads: "_Why per-capita vs raw counts? Click here to find out >>_" Best practices could be its own piece that builds over time. Any future visualizations that require justification can link to a section within best practices.

covid-dash@2x

goleary:
@kevee I'd like to get your opinion on the file structure of our components that are meant to be used solely within the `dashboard` page.

I've already moved the common Chart components to `src/components/charts` and am working on converting them to use scss modules.

I _prefer_ to keep (non-common) components close to the page within which they are used which is why they currently live in `src/pages/dashboard` instead of `src/components/pages/dashboard`.

That said I don't feel super strongly and if y'all would prefer that we bring _everything_ into the same file structure as the rest of the project we can make that happen.

kevee:
I think it's appropriate for page-only components to be scoped to their page, so what you said is fine with me.

src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -0,0 +1,292 @@
...
+  parseDate,
+  totalColor,
+  positiveColor,
+} from '../../utilities/visualization'
+
+import './dashboard.scss'
+
+// these come from this google spreadsheet owned by Júlia Ledur
+// https://docs.google.com/spreadsheets/d/1mD_NhlJR1fM2Pv_pY8YixUrX2p2F8rAE0xPTtsTJOiM/edit#gid=0
+const stayAtHomeOrders = {
```
kevee, line 21: This should really be in a separate file in `/src/data` or in our API eventually.

src/pages/dashboard/_UsCumulativeDeathsContainer.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
```
kevee, line 1: In general, let's try and rename these to just kebab-case-files. Also, the underscore should be removed.

src/components/charts/area-chart.js
```diff
@@ -0,0 +1,185 @@
...
+  marginRight: 0,
+  marginTop: 0,
+  xTicks: 5,
+  yMax: null,
+  yTicks: 4,
+  yFormat: null,
+  showTicks: true,
+}
+
+AreaChart.propTypes = {
```
kevee, line 160: Thank you for being a good citizen and using prop types!

src/components/charts/area-chart.scss
```diff
@@ -0,0 +1,10 @@
+.area-chart__y-tick-label {
```
kevee, line 1: I know SCSS modules are on the horizon, in that case you can remove the `.area-chart__` stuff.

src/utilities/visualization.js
```diff
@@ -0,0 +1,73 @@
+import { format } from 'd3-format'
+import { timeFormat, timeParse } from 'd3-time-format'
+
+export const formatDate = timeFormat('%b. %e')
+export const formatNumber = format(',.0f')
+export const parseDate = timeParse('%Y%m%d')
+
+export const getStateName = abbr => {
```
kevee, line 8: These can be pulled from the GraphQL database using staticQuery.

src/utilities/visualization.js
```diff
@@ -0,0 +1,73 @@
...
+    VT: 'Vermont',
+    WA: 'Washington',
+    WI: 'Wisconsin',
+    WV: 'West Virginia',
+    WY: 'Wyoming',
+  }
+  return names[abbr] || abbr
+}
+
+export const totalColor = '#585BC1'
```
kevee, line 70: Let's move these to the `colors.scss` file.

src/utilities/visualization.js
```diff
@@ -0,0 +1,73 @@
+import { format } from 'd3-format'
+import { timeFormat, timeParse } from 'd3-time-format'
+
+export const formatDate = timeFormat('%b. %e')
```
kevee, line 4: There's a format time and format number component in `/src/components/common/format` - maybe we can reuse them? I'm trying to keep all the date format stuff in one place because it lets us maintain a consistent style.

src/pages/dashboard/state-cumulative-deaths-container.scss
```diff
@@ -0,0 +1,29 @@
+@import '../../scss/breakpoints.scss';
+section.state-cumulative-death-container {
```
kevee, line 2: You will need to remove the tag selector when we move to modules.

PR #432, opened 2020-04-08
converted US Area chart to use Gatsby data

goleary:
first PR converting a chart to using Gatsby data instead of fetching from API.
You can see how much faster the top chart loads than everything below.

For using graphql queries within a component (instead of a page) you must use a `<StaticQuery>` component.  More info can be found [here](https://www.gatsbyjs.org/docs/static-query/).

@kevee would like you to take a look and make sure we've got your vote of confidence before converting the rest of the dashboard using this pattern.

### DIFF
--- a/src/pages/dashboard/_UsAreaChartContainer.js
+++ b/src/pages/dashboard/_UsAreaChartContainer.js
@@ -1,0 +1,55 @@
+import React from 'react'
+import { StaticQuery, graphql } from 'gatsby'
+import AreaChart from './charts/_AreaChart'
+import { parseDate } from './_util'
+
+export default function UsAreaChartContainer() {
+  const transformData = data => {
+    const transformedData = data.allCovidUsDaily.edges
+      .map(({ node }) => [
+        {
+          date: parseDate(node.date),
+          label: 'Total',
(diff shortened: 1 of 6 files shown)


src/pages/dashboard/_UsAreaChartContainer.js
```diff
@@ -0,0 +1,55 @@
...
+          date: parseDate(node.date),
+          label: 'Total',
+          value: node.totalTestResults,
+        },
+        { date: parseDate(node.date), label: 'Positive', value: node.positive },
+      ])
+      .flat()
+    return transformedData
+  }
+  const query = graphql`
```
kevee, line 20: This could also be shorter with [useStaticQuery](https://www.gatsbyjs.org/docs/use-static-query/#basic-example)

```javascript
const data = useStaticQuery(graphql`... your query ...`)

return (
<AreaChart data={data} ..../>
)
```

goleary: Ooh nice, I love hooks, I'll switching over to using this.

goleary:
i'm gonna go ahead and merge this so other have it as an example if they want to convert any other components.

Issue #353, opened 2020-04-02
change tooltip table style 


julialedur:
I suggest adding colored underlines to the variables on the table so it's easier to read from a visual perspective. we can also get rid of that gray background (sorry, that was a bad idea!) like this: 

<img width="606" alt="Screen Shot 2020-04-02 at 1 32 22 AM" src="https://user-images.githubusercontent.com/39416238/78213985-1c719e00-7482-11ea-8b4c-d1a4b9334f0b.png">

let's also take the font of the table a bit down, so there's a visual hierarchy (state name is bigger than table info) 

also, it looks like there's a right margin on the tooltip now:

<img width="47" alt="Screen Shot 2020-04-02 at 1 35 37 AM" src="https://user-images.githubusercontent.com/39416238/78214080-50e55a00-7482-11ea-9465-02217b6c811f.png">

julialedur:
we need to add some margin around this: 

<img width="274" alt="Screen Shot 2020-04-07 at 3 13 50 PM" src="https://user-images.githubusercontent.com/39416238/78709888-6cb58980-78e2-11ea-9cea-6044b6ac83ff.png">

friej715:
Holler if this looks good and I'll make a PR. 
![image](https://user-images.githubusercontent.com/1028403/78740169-0b160f00-7924-11ea-8327-e0884dfc642e.png)

friej715:
Oh, never mind, saw the PR above.

jeremiak:
Thanks though @friej715 !!

nselikoff:
Sorry @friej715 wires crossed

PR #435, opened 2020-04-08
fix(visualization): Changed tooltip style for map

nselikoff:
Note that the positioning of the tooltip was also changed in this commit. Previously, the tooltip was `position: fixed`, which caused the tooltip to not scroll with the page. Switched it to absolute + added a new relative div to position it consistently across the choropleth and bubble maps.

fixes issue #353

### DIFF
--- a/src/pages/dashboard/charts/_Map.js
+++ b/src/pages/dashboard/charts/_Map.js
@@ -118,25 +118,27 @@ export default function Map({
             <BubbleLegend data={data} r={r} maxValue={maxValue} />
           ))}
       </div>
-      <svg width={width} height={height}>
-        {data && (
-          <>
-            {!useChoropleth && (
-              <Bubbles geoJson={data} getValue={getValue} r={r} />
-            )}
-            <States
-              geoJson={data}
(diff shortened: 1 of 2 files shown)


src/pages/dashboard/charts/_Map.js
```diff
@@ -269,33 +274,58 @@ const Tooltip = ({ hoveredState, currentDate, getValue }) => {
...
+          <span className="date">{formatDate(parseDate(currentDate))}</span>
+        </caption>
         <thead>
           <tr>
-            <td colSpan="3">
-              {d.properties.NAME}
-              <br />
-              <span className="date">{formatDate(parseDate(currentDate))}</span>
-            </td>
+            <th scope="col">Metric</th>
```
nselikoff, line 284: Also wanted to note why I added "Metric" to this previously empty cell. Per https://webaim.org/techniques/tables/data, 

> Table headers should never be empty. This is particularly of concern for the top-left cell of some tables

Also and more practically, when I was switching this from `<td />` to `<th />` there was a linting rule complaining at me :)

PR #436, opened 2020-04-08
Use GraphQL query for small multiples charts

jeremiak:
Uses the pattern established in #432 to query the data from the GraphQL API for the small multiples charts, instead of fetching data client side.


jeremiak:
@goleary I'm going to merge this, but if you have a chance to review and notice anything that should be changed I'll pick it up in another PR.

PR #443, opened 2020-04-08
Break out dashboard into section containers that can fetch data

jeremiak:
Each section can use the GraphQL queries to get the data it needs and put it into charts.


src/pages/dashboard/_UsPositiveAndTotalTestsContainer.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+import { graphql, useStaticQuery } from 'gatsby'
+import AreaChart from './charts/_AreaChart'
+import { parseDate } from './_util'
+
+export default function UsAreaChartContainer() {
+  const transformData = data => {
```
goleary, line 7: i would put `useStaticQuery` first for improved readability/code flow.

src/pages/dashboard/_UsPositiveAndTotalTestsContainer.js
```diff
@@ -28,6 +15,18 @@ export default function UsAreaChartContainer() {
       }
     }
   `)
+  const transformedData = useMemo(() => {
```
goleary, line 18: technically `data` will NEVER change so using `useMemo` here is unnecessary.
The data comes in at build time once and is hard coded in the built file.

Issue #349, opened 2020-04-02
adjust legend of choropleth 


julialedur:
let's use these colors for the choropleth, so that the yellow we're using across the page for positive tests matches the map scale: 

<img width="349" alt="Screen Shot 2020-04-01 at 9 52 43 PM" src="https://user-images.githubusercontent.com/39416238/78202865-6bf4a180-7463-11ea-99b2-dd478ae65ff1.png">

here are the hex codes: 
#FCF9EB
#FBE8A9
#F6CE7A
#F3B05D
#E2894E
#C66B3E
#924F34
#753C2D

let's also make the buttons font-size a bit smaller and add "map" after choropleth 

let's delete "per million residents" from the choropleth legend and add it here instead: 

<img width="538" alt="Screen Shot 2020-04-02 at 12 57 18 AM" src="https://user-images.githubusercontent.com/39416238/78212180-fe556f00-747c-11ea-942c-2f2f891afb80.png">

I'm opening another issue for dropdown styling

goleary:
@julialedur we got feedback on slack from several people that have no idea what 'choropleth' means. they suggested using 'state map' as the text for the toggle. do you have any ideas for more user friendly names to put in that?

Also are you satisfied with the colors present for 'deaths' & 'total test'?

julialedur:
yeah, i guess that makes sense! let's talk about it tomorrow on slack :) 
i talked to jason about the colors today and he's gonna give us new hex codes for the gray and purple scales. orangy is still this one: 

#FCF9EB
#FBE8A9
#F6CE7A
#F3B05D
#E2894E
#C66B3E
#924F34
#753C2D

vijaysharwar:
I can work on these issues.

vijaysharwar:
@julialedur PR opened for fixes https://github.com/COVID19Tracking/website/pull/445

Let me know if there are new hex codes for the grey and purple schemes. Can add to this PR, or break into separate issue if you are still deciding on new colors. Same if "choropleth" term needs to be changed.

vijaysharwar:
@julialedur PR merged https://github.com/COVID19Tracking/website/pull/445 by @jeremiak. But let me know about those other potential changes.

julialedur:
hey @vijaysharwar and @goleary! 
we got new colors from Jason today: 

*main colors*

purple (total tests): #575AAD
orange (positive tests): #FFAD4A
gray (deaths): #4C5559 


*scales*

Honey:
#6E2F1F
[...]

PR #445, opened 2020-04-08
Adjust legend of choropleth

vijaysharwar:
- Updates orange color scheme
- Decreases toggle font size, updates "Choropleth" to "Chropleth Map"
- Moves/rewords "per one million people" text

Fixes COVID19Tracking/website#349.

### New layout:

Screen Shot 2020-04-08 at 1 41 44 PM

### DIFF
--- a/src/pages/dashboard/_MapContainer.js
+++ b/src/pages/dashboard/_MapContainer.js
@@ -146,7 +146,7 @@ const MapContainer = () => {
         tabIndex={0}
       >
         <span className={useChoropleth ? '' : 'active'}>Bubble Map</span>
-        <span className={useChoropleth ? 'active' : ''}>Choropleth</span>
+        <span className={useChoropleth ? 'active' : ''}>Choropleth Map</span>
       </div>
       <div id="map-dek">
         <h2>{formatDate(parseDate(currentDate))}</h2>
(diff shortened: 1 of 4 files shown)


src/pages/dashboard/_MapContainer.js
```diff
@@ -146,7 +146,7 @@ const MapContainer = () => {
         tabIndex={0}
       >
         <span className={useChoropleth ? '' : 'active'}>Bubble Map</span>
-        <span className={useChoropleth ? 'active' : ''}>Choropleth</span>
+        <span className={useChoropleth ? 'active' : ''}>Choropleth Map</span>
```
jeremiak, line 149: Good catch, we're gonna use a better word here than "choropleth" since most folks have no idea what that is

src/pages/dashboard/charts/_ChoroLegend.js
```diff
@@ -45,17 +45,14 @@ export default function ChoroLegend({
     </g>
   )
   return (
-    <>
-      <h5 style={{ marginBottom: 0 }}>per million residents</h5>
```
jeremiak, line 49: Dig removing the text, though we might want to make this a prop since all legends are going to need it

PR #446, opened 2020-04-08
feat(dashboard): converted map to use gatsby data

goleary:
Map Container now gets the data from gatsby and passes it to the `Map` component.

Question for you @kevee:

The dashboard container also pulls geojson data from `/src/pages/dashboard/data/_state-populations.js` which it joins the state level data from `allCovidStateDaily` to.

My assumption is that because this file/data is imported this is performed at compile/build time (this appears to be the case from looking at how much faster the map loads) so is already benefiting from Gatsby.

That said, part of me feels like this data should instead live within gatsby so that it conforms to the rest of the project and so that others can use this data if interested. 

Do you think it's worth me adding this to the gatsby data layer?


kevee:
One hundred percent. That should live in `/src/data/....` and you can use `gatsby-source-filesystem` to query it.

goleary:
decided to leave the JSON importing as is due to headache desribed [here](https://www.gatsbyjs.org/packages/gatsby-transformer-json/#troubleshooting)

Issue #430, opened 2020-04-07
small multiples axes and size 


julialedur:
*x-axis*
we need to fix the x-axis for the small multiples. the last ones a weird again :/ 

<img width="1296" alt="Screen Shot 2020-04-07 at 3 20 24 PM" src="https://user-images.githubusercontent.com/39416238/78710463-6673dd00-78e3-11ea-82c5-006a2edd1f10.png">

*y-axis*
let's add some breathing space between grid lines and tick values 
delete all values except for the first of every row 
change all grid lines to #b2bbbf 

*size*
let's also try to put 8 states per row? the graphic is very long now and it will be easier to compare the states with more per row. they can also be 50% tinier

*NY chart*
the spike in NY is cut off 

<img width="330" alt="Screen Shot 2020-04-07 at 3 29 33 PM" src="https://user-images.githubusercontent.com/39416238/78711198-9a033700-78e4-11ea-948e-d13e75f044f8.png">

vijaysharwar:
Working on this.

vijaysharwar:
@julialedur @jeremiak I'm not sure about the proposed solutions. I actually think the individual charts need to be _larger_, not smaller.

The biggest issue is the difference in scale between a state like NY and the majority other states with far lower testing numbers. If we increase the y-axis scaling to accommodate NY's overflow, you won't see *anything* on the vast majority of charts, because their test numbers are too low by comparison.

We could have a different X/Y scale per state, but that would make comparisons very difficult.

I think a better starting place might be one, taller state chart per row, which can solve issues of both X and Y axes (and work better on small/mobile screens).

From there, could consider additional improvements:
- Show/hide toggle on each chart (plus "show/hide all") to optionally reduce number of charts, compare smaller number at a time
- Toggles between linear and log scaling on y-axis; this would shrink the visual discrepancy between large and small number of tests (example at [https://www.coronaviruschart.com/](https://www.coronaviruschart.com/))

jeremiak:
Ya, it definitely presents some challenges. I think we want to do something like this, where one or two of the big outliers just break through their chart so that they're still useful and the rest of the charts are still intelligible.

<img width="1199" alt="Screen Shot 2020-04-08 at 2 31 57 PM" src="https://user-images.githubusercontent.com/780941/78820452-d21e7e80-79a5-11ea-8bca-3eb7be0070ba.png">

[...]

PR #449, opened 2020-04-08
fix(charts): reformat small area charts for legibility

vijaysharwar:
Trying a few things here to resolve various x/y axis and comparative chart
layout issues as described in #430:

- Made all area charts wider aspect ratio, to improve y-axis scaling with
  outlier states that have high number of tests.
- Switched to viewBox attribute for chart SVGs to allow responsive scaling
  to their containers
- On smaller screens < 960px, go to single column. > 960px go to two column
  format, looking for sweet spot of legibility and comparative analysis btwn
  states with high vs low number of tests
- Tweaked tick text font size, improve number of ticks consistency
- Adjusted the small charts legend for better responsive layout
- Updated grid line colors to #b2bbbf

### Small screen:
Screen Shot 2020-04-08 at 5 31 37 PM

### Large screen:
Screen Shot 2020-04-08 at 5 31 57 PM


src/pages/dashboard/_SmallMultiplesContainer.js
```diff
@@ -105,11 +105,12 @@ const SmallMultiplesContainer = () => {
...
-    if (!data[1] && !data[0]) return
-    // eslint-disable-next-line consistent-return
-    return max(data[1].values, d => calculateTotal(d))
-  }, [data.length])
+
+  const maxStateTests = useMemo(() => {
+    return max(data, stateData => {
+      return max(stateData.values, d => calculateTotal(d))
+    })
+  })
```
vijaysharwar, line 113: @jeremiak I revised this logic determining the max Y scaling. Unsure why code was using the first two only originally, but maybe I'm missing some context.

julialedur:
pasting this here icymi. 
a few other things: 
- legend on mobile needs to be smaller 
- chart should still be a small multiples on mobile 

@vijaysharwar though this is a fairly exploratory viz, there are certain aspects that should be set by us, like number of columns. otherwise it just gets too confusing and impossible for us to get our point across. let's implement your toggle button for per capita numbers. that way we can show:
- how NY's raw numbers are incredibly high 
- the comparison between states and how normalizing for population makes a difference

they should be tiny tiny though as this is not supposed to be a detailed zoomed in graphic. like this: 

example

 it will be clearer once we add a tooltip.

vijaysharwar:
@julialedur I can update the PR, it would look something like this:

state-charts

At that scale, the numbers are completely unreadable. And you can't see the amount of testing at all for the vast majority of states.

Maybe that can be improved by putting a larger chart on the individual state page (accessed by the "See all data from state" text links, though.

On that note, those "See all data from state" links are now overwhelming/repetitive. Should we make the text pithier and/or font size smaller, if we're making the charts this tiny?

julialedur:
can we do 7 per row and normalize by population? that way the rest of the states won't be so low and we don't have a lost one at the end

vijaysharwar:
I think normalizing by population should be a separate issue.

vijaysharwar:
@julialedur I think 7 columns only works on larger screens. I've handled this responsively in https://github.com/COVID19Tracking/website/pull/449/commits/f2fcc08f8b7cf61e5ab55f2c9a3f77b480cfd143.

I also tried making the legend smaller, but if it still doesn't work please provide a mockup. Having trouble fitting that much info and keeping it legible/accessible.

### <480px

Screen Shot 2020-04-08 at 7 07 51 PM

### 480-799px

Screen Shot 2020-04-08 at 7 09 15 PM

### 800-1199px

Screen Shot 2020-04-08 at 7 10 38 PM

### 1200px and above

Screen Shot 2020-04-08 at 7 11 14 PM

julialedur:
that looks nice! thanks for making it responsive too :) 

my thoughts: 
- keep only 100,000 and 200,000 ticks (it will remove noise) 
- we'll definitely need to normalize by pop. rn the chart doesn't tell much of a story (i'll create another issue for this) 
- let's change the last key in the legend to only "Stay-at-home order" and add a note below the chart that says "**Note:** Only statewide stay-at-home orders are included; Dates mark when the orders went into effect." a bit smaller than body text.

vijaysharwar:
@julialedur if they're this small, the text and ticks are meaningless. Should we just remove them altogether? (We can keep them for the larger US totals chart at top of dashboard)

vijaysharwar:
Here's where we're at with 17e3d49. I put the new "Note" above the legend, because otherwise it conflicts with the legend stickiness-on-scroll functionality.

Not sure what you mean with "make mobile version collapsible", but sounds like it should be a separate issue. With many contributors, best to keep PRs short-lived and concise to avoid merge  conflicts.

Screen Shot 2020-04-09 at 12 31 33 PM

julialedur:
- sure, let's remove the tick labels and the grid. then we'll add a tooltip with this info.

- let's move the note to below the chart, it doesn't make sense above it. if you put it outside the chart's div it won't mess with the scroll functionality. 

- i'll open an issue to deal with the collapsible mobile version

vijaysharwar:
@julialedur cool, how's this? 86f6e7c

Screen Shot 2020-04-09 at 2 38 05 PM

julialedur:
oh sorry, i meant all the way at the bottom, after the chart! 
and Only should be capitalized ;)

julialedur:
looks great otherwise! so much cleaner without grids and labels

vijaysharwar:
Cool :)

What about doing it with an asterisk, like a footnote? That way it's linked to the legend, and people are less likely to miss it.

legend-note

julialedur:
Great!

vijaysharwar:
Done in 1b05594. I think this should be good to merge. cc @jeremiak @goleary @kevee

src/pages/dashboard/_SmallMultiplesContainer.js
goleary: we originally had NY going off the graph to demonstrate how bad things have gotten there.

PR #450, opened 2020-04-08
fix(choropleth): update color schemes

vijaysharwar:
Uses new custom color schemes for total test results and death maps.

### Positive

choropleth-positive

### Totals

choropleth-total

### Deaths

choropleth-deaths

Fixes COVIDTracking/website#349


vijaysharwar:
My only question is on the grey scheme. We're losing contrast with the state boundary lines. Compare to what was there previously:

Screen Shot 2020-04-08 at 6 02 42 PM

cc @julialedur

julialedur:
i agree that the grays are lacking a bit of contrast on the map. 
I'll ask Jason to tweak the colors a bit. 

about state lines, could we make them white just on the gray map? 
thanks!

vijaysharwar:
@julialedur done in https://github.com/COVID19Tracking/website/pull/462.

Issue #429, opened 2020-04-07
weird blue box on button click


julialedur:
A weird blue box show up whenever you click buttons on the page: 

<img width="1266" alt="Screen Shot 2020-04-07 at 3 17 00 PM" src="https://user-images.githubusercontent.com/39416238/78710253-02511900-78e3-11ea-8419-c78d9af7b54f.png">

goleary:
I think this is probably imposed by global styles for the entire website. It should however only cover the visible toggle instead of the whole line.

AaronMullan:
Hold on, pull request coming.

AaronMullan:
PR is in.

PR #451, opened 2020-04-08
fix(charts): remove focus from container, re: #429

AaronMullan:


vijaysharwar:
@AaronMullan @goleary I know this is merged, but I would recommend reverting this for accessibility. Without `outline`, if you tab navigate to this UI, a keyboard-only user can't tell where the focus is.

Further reading:
https://webaim.org/techniques/keyboard/
https://webaim.org/blog/plague-of-outline-0/
https://www.w3.org/TR/WCAG20-TECHS/G202.html

AaronMullan:
I think the main thing people were not liking was the fact that the whole div gets focus, not just the  button.  I'm trying to look at a way of making only the button get focus.

AaronMullan:
I can just put it in another div.  Here it's in the middle, which I don't mind, but I guess that's not the design.

Screen Shot 2020-04-10 at 4 47 25 PM

AaronMullan:
Or float the button to the left:

Screen Shot 2020-04-10 at 4 52 14 PM

vijaysharwar:
@AaronMullan we can do it by [adjusting the width](https://github.com/COVID19Tracking/website/blob/dashboard-gatsby/src/pages/dashboard/map-container.scss#L72) perhaps.

Note changes I'm making in another PR, though:
- Changing the ID to a class: https://github.com/COVID19Tracking/website/pull/494/files#diff-06cd32530420ab6e5e788a7f09e1f4a3R70
- Setting a `max-width` attribute on my new toggle to restrict the size: https://github.com/COVID19Tracking/website/pull/494/files#diff-d99b5a549ac78593a54c2878d296386cR62-R66

Might be easiest to revert this first, merge PR above, then fix the width of the original map toggle last.

Also note that the `outline` style change is very brute force, and removes outline for all elements in `.container` (including the global top nav).

AaronMullan:
Anything that is the simplest path forward sounds good to me. I hadn't realized focus was important for those reasons. The solution I came up with was just to make another div as the container for the button, give the container a width of 100%, and take the 100% width off the button div.  Once your PR's are merged one of us could do that.

goleary:
a PR from either of you reverting it to the accessible state will be welcomed.

AaronMullan:
PR #500 addresses this.

Issue #398, opened 2020-03-26
API: Add geo and population stats


webmasterkai:
I need a source of the information. Is that something we are maintaining a copy of or are we locating an official source that can be directly pulled from?

gitjeff05:
Perhaps this is not part of this PR, but what kind of standardized date format is "Mar. 26, 2020?"

Also, couldn't the date updated just be _one_ field at the top instead of on each element? Would these values ever be updated individually?

Issue #260, opened 2020-03-30
Add state population data to the website


preislerd:
As a consumer of the website, when I see the test and outcome numbers, I always ask myself, what is the population we are talking about.  It would be great to have a population number next to the state names, or below the state boxes.  And the same for US total.  Ideally it would be a trusted static data source that provides context for the testing data (like Census data).

I've asked @metasj to look into collecting and adding the data to the repo.

Other thoughts --
1. It should be on the website as described above.
2. It should be available in a public sheet -- probably in the sheet that has "todays" or "current data".
3. It should be available in the API.

metasj:
I added mid-2019 census population counts to the States [worksheet](https://docs.google.com/spreadsheets/d/1MvvbHfnjF67GnYUDJJiNYUmGco5KQ9PW0ZRnEP9ndlU/edit#gid=1208387230); we should be able to easily draw from that into the site.  I'll post the source file to associated-data for safekeeping.

PR #455, opened 2020-04-09
feat(api, design): Added population data to the website and the api.

jgulian:
This pull request adds population data to the api/website as suggested in #260 , and #398. It adds population numbers to the state list, and state pages. It also adds the ability to add static json files to the api as shown with build/static/population.json.


jgulian:
I know putting population data in the api was of some concern. If that is an issue, it would be really easy to remove that, and just put populations on the website.

kevee:
Hi, @jgulian, this is on my radar and I'm getting feedback from our data team on how to best handle this. Thanks so much for your PR and hold tight!

PR #459, opened 2020-04-09
feat: Centralize formatting functions into components, also, date formatting

### LINKED ISSUE #458: fix: Clean up default date format
Our current default date format for history tables is is `dd LLL yyyy ccc`, which yields: 

08 Apr 2020 Wed

Ugh

I would suggest changing this to ([token table here](https://moment.github.io/luxon/docs/manual/formatting.html#table-of-tokens)):

`ccc LLLL d yyyy`
`[abbreviated day of week] [full month] [day of month] [year]`
Mon April 8 2020

I'm not sure why we have day of the week, but at the same time it *is* helpful when saying "but look what this number was last Wednesday." I have now exhausted the five minutes I have allowed myself to think about this.

kevee:
Fixes #458 

This PR:

- Moves the `formatDate` and `thousands` utilities to regular React components
- Makes numbers default to 'N/A' if not available so you don't have to have ternary operators all over the place
- Uses a more readable default date format

### DIFF
--- a/src/components/common/format.js
+++ b/src/components/common/format.js
@@ -1,0 +1,17 @@
+import React from 'react'
+import { DateTime } from 'luxon'
+
+const FormatDate = ({ date, format = 'ccc LLLL d yyyy' }) => {
+  if (typeof date === 'undefined') {
+    return null
+  }
+  return <>{DateTime.fromISO(date).toFormat(format)}</>
+}
+
+const FormatNumber = ({ number }) => (
+  <>
+    {number ? number.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',') : 'N/A'}
+  </>
+)
+
+export { FormatDate, FormatNumber }
--- a/src/components/common/format.js
+++ b/src/components/common/format.js
@@ -9,9 +9,7 @@ const FormatDate = ({ date, format = 'ccc LLLL d yyyy' }) => {
 }
 
 const FormatNumber = ({ number }) => (
-  <>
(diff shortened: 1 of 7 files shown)


src/pages/data/us-daily.js
```diff
@@ -38,17 +39,29 @@ const ContentPage = ({ data }) => (
         {data.allCovidUsDaily.edges.map(({ node }) => (
           <tr>
             <td>
-              {DateTime.fromFormat(node.date.toString(), 'yyyyMMdd').toFormat(
-                'dd LLL yyyy ccc',
-              )}
+              <FormatDate
+                date={DateTime.fromFormat(node.date.toString(), 'yyyyMMdd')}
```
schwartzadev, line 43: Is there a reason why this date format is different from the default one?

src/components/common/format.js
```diff
@@ -0,0 +1,17 @@
...
+const FormatDate = ({ date, format = 'ccc LLLL d yyyy' }) => {
+  if (typeof date === 'undefined') {
+    return null
+  }
+  return <>{DateTime.fromISO(date).toFormat(format)}</>
+}
+
+const FormatNumber = ({ number }) => (
+  <>
+    {number ? number.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',') : 'N/A'}
```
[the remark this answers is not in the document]
kevee: I think we were not using t`oLocaleString()` because it throws errors when number is null. I'm going to add more logic to handle all edge cases then we can use `toLocaleString()`

src/pages/data/us-daily.js
kevee: Unsure, checking.

PR #461, opened 2020-04-09
Tweaking the color naming scheme (closes #460, closes #465)

### LINKED ISSUE #460: Adopt a `font-weight`-like naming system for our color palette
> Let us know what the design issue is.

Per @jasonsantamaria’s suggestion in https://github.com/COVID19Tracking/website/issues/440#issuecomment-611524394, let’s use a `font weight`-esque numeric system for our color “weights”! Here’s Jason’s suggestion:

>  I've had good luck with using the same convention as CSS font-weight decorations (400, 600, 900, etc) to allow for some space to expand.
> 
> So instead of $color-plum-1 you would do $color-plum-100, then $color-plum-200, $color-plum-300, and so on. That way later if you need to introduce an additional step somewhere in the middle, you could do $color-plum-250 and not have to renumber everything.

> Screenshots are helpful!

no screenshots its just css times

### LINKED ISSUE #465: Redux slate gray palette
The data folks are having some trouble getting enough contrast across the `slate` palette (as mentioned in #450). It's the only color that has 10 steps, and I'm thinking we should bring it in line with the others at 8 steps. That will make sure we have a little more contrast that they need for choropleth maps.

I'm not sure how many of the existing  `slate` colors are in use, so this may or may not be hairy to change. I'm happy to consult translating any gray use over. I imagine the potential problem areas will be any at the very light or very dark ends of the spectrum. 

The new `slate` range would be:

```

Slate:
#313638
#4C5153
#676B6D
#828688
#9CA1A2
#B7BCBD
#D2D6D7
#EDF1F2
```

Color palette Redux

beep:
As suggested by @jasonsantamaria, this PR proposes we adopt a `font-weight`-like naming scheme for our color variables. Specifically:

- Current color variables increment by steps of 100. (`$color-plum-100`, $color-plum-200`, and so on)
- Darker shades of each color—think “heavier weights”—are placed toward the upper range of each variable set. (`$color-honey-200` is lighter than `$color-honey-300`; `$color-honey-800` is darker than both)

Feedback/discussion welcome!

### DIFF
--- a/src/components/common/detail-text.module.scss
+++ b/src/components/common/detail-text.module.scss
@@ -3,5 +3,5 @@
 .detail-text {
   font-size: 0.8rem;
   margin-bottom: 0.3rem;
-  color: $color-slate-3;
+  color: $color-slate-800;
 }
--- a/src/components/common/detail-text.module.scss
+++ b/src/components/common/detail-text.module.scss
@@ -3,5 +3,5 @@
 .detail-text {
   font-size: 0.8rem;
   margin-bottom: 0.3rem;
-  color: $color-slate-800;
+  color: $color-slate-700;
 }
--- a/src/components/layout/footer.module.scss
+++ b/src/components/layout/footer.module.scss
@@ -46,7 +46,7 @@
 
         &:hover,
         &:focus {
-          color: $color-plum-7;
+          color: $color-plum-200;
         }
       }
     }
--- a/src/components/layout/header.module.scss
+++ b/src/components/layout/header.module.scss
@@ -108,11 +108,11 @@
         }
         a:hover,
         a:focus {
-          color: $color-blueberry-5;
+          color: $color-blueberry-100;
         }
         a:hover,
         a[aria-current] {
-          border-bottom: 3px solid $color-blueberry-4;
+          border-bottom: 3px solid $color-blueberry-200;
           padding-bottom: 2px;
         }
         a[aria-current] {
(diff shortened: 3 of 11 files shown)


beep:
(Hold off on reviewing/merging for a bit; gonna see if I can work in a fix for #465.)

vijaysharwar:
@beep I didn't realize the colors were to be centralized when I updated the maps in https://github.com/COVID19Tracking/website/pull/450. Might want to DRY things up in [_Map.js](https://github.com/vijaysharwar/website/blob/8acf4584440d302b0dac9b1b4577599d932ae6d4/src/pages/dashboard/charts/_Map.js#L61-L92).

beep:
@vijaysharwar: Agreed! If those color values need to live in JS, you should be able to import those values directly from colors.scss if you want. (à la https://github.com/COVID19Tracking/website/blob/master/src/stories/2-components/03-Colors.stories.js#L2)

(If you want to discuss a bit more, happy to take this discussion to Slack or to another issue!)

beep:
Got [a verbal 👍 from @jasonsantamaria in Slack](https://covid-tracking.slack.com/archives/C010N48RY73/p1586461233160900?thread_ts=1586458764.156600&cid=C010N48RY73), so I think this passes design muster! If anyone else has any feedback, I’d love to hear it!

vijaysharwar:
@beep I've not yet been invited to the Slack, but if you want to make an issue once this is merged, I can tackle the refactor.

beep:
@vijaysharwar done, and thank you! https://github.com/COVID19Tracking/website/issues/469

Issue #463, opened 2020-04-09
add "see all data for state" names 


julialedur:
@beep made a good suggestion to make this chart more accessible: 
- let's use "See all data for [state name]" instead of "See all data for state" 

so we'd have "See all data for New York", "See all data for New Jersey", etc.

AaronMullan:
Pull Request coming soon.

jeremiak:
Done in the above PR

PR #467, opened 2020-04-09
fix(charts): make links have state names #463

AaronMullan:


AaronMullan:
It's weird, the 'Resolve conflicts' button is greyed out.  And there's really only one line of one file that's changed, so I'm confused where the conflict is coming from.

goleary:
@AaronMullan I've seen some weird issue like this before.  Just merge the latest `dashboard-gatsby` to your branch from the cli and then push the result.  There may not even be any conflict to merge...

AaronMullan:
OK, that seemed to work!

goleary:
cool, thanks!

PR #468, opened 2020-04-09
Add basic CDC chart with data from GraphQL API

jeremiak:
Still needs some styling work but good enough for review/merge and we can break out other issues to fix the small things.

Screen Shot 2020-04-09 at 3 48 28 PM

### DIFF
--- a/src/pages/dashboard/_CDCComparisonContainer.js
+++ b/src/pages/dashboard/_CDCComparisonContainer.js
@@ -1,9 +1,18 @@
-import { sum } from 'd3-array'
-import React, { useMemo } from 'react'
+import { max, sum } from 'd3-array'
+import { timeParse } from 'd3-time-format'
+import React from 'react'
 import { graphql, useStaticQuery } from 'gatsby'
 
 import BarChart from './charts/_BarChart'
-import { calculateTotal, formatNumber, parseDate } from './_utils'
+import { formatNumber, parseDate } from './_utils'
+
+const parseCdcDate = timeParse('%m/%d/%Y')
+
+const sortChronologically = (a, b) => {
+  if (a.date > b.date) return 1
+  if (a.date < b.date) return -1
+  return 0
+}
 
 export default function CDCComparisonContainer() {
   const query = useStaticQuery(graphql`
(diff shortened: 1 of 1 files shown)


src/pages/dashboard/_CDCComparisonContainer.js
```diff
@@ -1,64 +1,101 @@
...
-          fill="#585BC1"
-          height={400}
-          marginBottom={40}
-          marginLeft={80}
-          marginRight={10}
-          marginTop={10}
-          xTicks={2}
-          width={200}
-        />
+      <div style={{ display: 'flex', flexGrow: 1, width: '50%' }}>
```
jeremiak, line 71: All this inline style business is going to have to change, but just wanted to get this up quickly

src/pages/dashboard/_CDCComparisonContainer.js
```diff
@@ -1,64 +1,101 @@
...
+  const data = query.allCovidUsDaily.nodes
+    .map(node => [
+      {
+        date: parseDate(node.date),
+        value: node.totalTestResultsIncrease,
+      },
+    ])
+    .flat()
+    .sort(sortChronologically)
+  const cdcCumulativeTotal = sum(cdcData, d => d.value)
```
jeremiak, line 54: My understanding of the CDC data is that its all daily and none of it is cumulative, so we should be able to create a sum like this

src/pages/dashboard/_CDCComparisonContainer.js
```diff
@@ -1,64 +1,99 @@
...
-    return nodes.sort((a, b) => {
-      if (a.date > b.date) return 1
-      if (a.date < b.date) return -1
-      return 0
+  const cdcData = query.allCdcDaily.nodes
+    .map(node => {
+      const date = parseCdcDate(`${node.dateCollected}/2020`)
+      return {
+        date,
+        value: +node.dailyTotal,
```
goleary, line 41: what does the plus here denote?

jeremiak: It coerces the value to an integer

PR #473, opened 2020-04-10
feat(dashboard): add charts to storybook

goleary:
I didn't make much of any modifications to the components themselves.

The AreaChart & the BarChart now show up in storybook.  I also added PropTypes for both to make it clear what props are required and what types they should be.

We could probably do with some further cleanup to make these more friendly for use elsewhere.

The AreaChart specifically doesn't respect the height & width passed, it just seems to use them for proportions & relies on the container element to size it. (i'll add a issue on this one).

The map I'll have to think about a bit more (assuming anyone wants to use it elsewhere). It is a bit more entangled with it's container component at the moment.

### DIFF
--- a/src/pages/dashboard/charts/_AreaChart.js
+++ b/src/pages/dashboard/charts/_AreaChart.js
@@ -8,6 +8,8 @@ import React from 'react'
 
 import { formatDate, formatNumber } from '../_utils'
 
+import './area-chart.scss'
+
 export default function AreaChart({
   annotations = null,
   data,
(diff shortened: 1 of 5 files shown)


goleary:
Awesome, glad you like them! I figured the request to restructure the files was coming sooner or later :)
I'll add a task on this.

goleary:
Awesome, glad you like them! I figured the request to restructure the files was coming sooner or later :)
I'll add a task on this.

Issue #452, opened 2020-04-09
Design refinements for the big ol text pages


kissane:
We have several long scrolly text pages that could use some love—they're looking real wiiiiide in big windows and just a little bit rougher in general than the data sections we asked the design team to focus on for the relaunch. I'd love to have some design attention and CSS nudges to get these shined up sometime. <3

Examples:

https://covidtracking.com/why-it-matters
https://covidtracking.com/help
https://covidtracking.com/about-data

kevee:
How about we wrap content pages in our ["narrow" container](https://covid19tracking.github.io/website-docs/?path=/story/layout--container-narrow)?

jasonlcrane:
I added a PR to use the narrow container. Any feedback is much appreciated.

beep:
@kissane Happy to take a crack at this tomorrow if they’re still unclaimed!

beep:
Here’s [a quick pass on what our text-heavy pages might look like](https://www.figma.com/file/fKUKsYMNw3NLSyIEOI1HKc/v2-refresh?node-id=0%3A14). And here’s a screenshot, if you’d prefer!

<img width="1440" alt="Content page" src="https://user-images.githubusercontent.com/124493/79138034-cd7d1000-7d81-11ea-9442-6a8a329edd04.png">

This mockup’s got two different subhed levels, numbered and bulleted list treatments, as well as a fair bit of body copy. Let me know if you’d like anything else included! But the main things here are:

- A baseline text size of 22px on widescreen, with a 1.5 line height.
- The content well’s width is capped at ~42rem.

Also, as I think @jasonlcrane noted on #474, it’d be _great_ to get a bit more spacing above and below the content area, just to create a bit more separation from the header & footer.

What do you folks think?

beep:
(Edited https://github.com/COVID19Tracking/website/issues/452#issuecomment-612976070, as I goofed when exporting the mockup. #goofs)

kissane:
Faaaaabulous ty @beep!

kissane:
> Also, as I think @jasonlcrane noted on #474, it’d be great to get a bit more spacing above and below the content area, just to create a bit more separation from the header & footer.

Yes please

toddparker:
~ chef kiss ~

schwartzadev:
I can tackle this today or tomorrow!

kevee:
Fixed in #535

PR #474, opened 2020-04-10
chore: Use narrow layout container for content pages

jasonlcrane:
This addresses #452, updating to use the narrow container for content pages on the site. I also felt like it needed some additional space between the header and start of the content on larger screens, but I defer to @beep and the design team.

Feedback welcome, especially since this is my first time poking my head into React.

### DIFF
--- a/src/components/common/container.js
+++ b/src/components/common/container.js
@@ -2,10 +2,9 @@ import React from 'react'
 import containerStyles from './container.module.scss'
 
 export default ({ children, narrow }) => (
-  <div
-    className={`container ${containerStyles.container} ${narrow &&
-      containerStyles.containerNarrow}`}
-  >
-    {children}
+  <div className={`container ${containerStyles.container}`}>
+    <div className={`${narrow && containerStyles.containerNarrow}`}>
+      {children}
+    </div>
   </div>
 )
--- a/src/components/common/container.js
+++ b/src/components/common/container.js
@@ -2,14 +2,13 @@ import React from 'react'
 import containerStyles from './container.module.scss'
 
 export default ({ children, narrow }) => {
-  if (narrow) {
-    return (
-      <div className={`container ${containerStyles.container}`}>
-        <div className={`${containerStyles.containerNarrow}`}>{children}</div>
-      </div>
-    )
-  }
   return (
-    <div className={`container ${containerStyles.container}`}>{children}</div>
+    <div className={`container ${containerStyles.container}`}>
+      {narrow ? (
+        <div className={`${containerStyles.containerNarrow}`}>{children}</div>
+      ) : (
+        children
+      )}
+    </div>
   )
 }
--- a/src/components/common/container.js
+++ b/src/components/common/container.js
@@ -4,7 +4,7 @@ import containerStyles from './container.module.scss'
 export default ({ children, narrow }) => (
   <div
     className={`container ${containerStyles.container} ${narrow &&
-      containerStyles.narrow}`}
+      containerStyles.containerNarrow}`}
   >
     {children}
   </div>
--- a/src/components/common/container.js
+++ b/src/components/common/container.js
@@ -5,9 +5,7 @@ export default ({ children, narrow }) => {
   if (narrow) {
     return (
       <div className={`container ${containerStyles.container}`}>
-        <div className={`${narrow && containerStyles.containerNarrow}`}>
-          {children}
-        </div>
+        <div className={`${containerStyles.containerNarrow}`}>{children}</div>
       </div>
     )
   }
--- a/src/components/common/container.js
+++ b/src/components/common/container.js
@@ -1,10 +1,17 @@
 import React from 'react'
 import containerStyles from './container.module.scss'
 
-export default ({ children, narrow }) => (
-  <div className={`container ${containerStyles.container}`}>
-    <div className={`${narrow && containerStyles.containerNarrow}`}>
-      {children}
-    </div>
-  </div>
-)
+export default ({ children, narrow }) => {
+  if (narrow) {
+    return (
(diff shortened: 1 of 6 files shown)


src/templates/content.js
```diff
@@ -8,6 +8,7 @@ const ContentPage = ({ data }) => {
     <Layout
       title={page.title}
       navigation={page.navigationGroup ? page.navigationGroup.pages : false}
+      narrow="true"
```
kevee, line 11: Props that are truthy should just be empty, so just change `narrow="true"` to `narrow`

```javascript
<Layout [stuff] narrow>
```

src/components/common/container.module.scss
```diff
@@ -17,5 +17,8 @@
   }
   &.container-narrow {
     max-width: 900px;
+    @media (min-width: $viewport-lg) {
```
kevee, line 20: I think we should leave the margin where it is.

src/templates/content.js
jasonlcrane: Thanks, makes sense. Updated.

src/components/common/container.module.scss
jasonlcrane: I updated to remove this.

jasonlcrane:
Thanks for the feedback, Kevin! Those fixes are made.

kissane:
Thanks very much for this @jasonlcrane! @beep when you have time next week could you give this a look for us? I like the width of the content much better, but I feel like our grid needs some design braining here.

schwartzadev:
Photos, for reference:

**current**:
image



**proposed**:
image

kevee:
I'm going to hold off for design review.

beep:
> @beep when you have time next week could you give this a look for us? I like the width of the content much better, but I feel like our grid needs some design braining here.

Absolutely, @kissane! I’ll take a look tomorrow morning.

src/components/common/container.module.scss
kevee: Thanks!

src/templates/content.js
kevee: 🙌

beep:
Hey folks! Just poking around a bit at this, and I think this is looking really promising. My main feedback is that the narrow layout probably shouldn’t be centered; ideally, we’d preserve that left-edge line, and pull the `max-width` in a bit. Here’s a quick in-browser mockup to show what I mean:

image

(This is with a `max-width` of 42rem, or ~670px. Something a bit narrower than 900px feels a bit more comfortable to me.)

What do you think? Does that seem doable? Happy to help with this, if I can.

jasonlcrane:
Great call to adjust the alignment and width @beep. I pushed up a new change to update for that. This changes the HTML structure slightly (adds a second container div) when using a narrow width, and will result in the narrow container not being able to be used to center narrow content on a page, at least as it is structured currently. That may have been the original intent, but it is not used for that purpose anywhere on the site that I am able to find. I updated the storybook to reflect this change as well.

src/components/common/container.js
```diff
@@ -1,11 +1,17 @@
...
-      containerStyles.narrow}`}
-  >
-    {children}
-  </div>
-)
+export default ({ children, narrow }) => {
+  if (narrow) {
+    return (
+      <div className={`container ${containerStyles.container}`}>
+        <div className={`${narrow && containerStyles.containerNarrow}`}>
```
kevee, line 8: This is within an if statement for narrow = true, so you can remove the `narrow &&` part.

kevee: Actually, I think you can just use this and not wrap it in an if statement:

```
<div className={`container ${containerStyles.container} ${narrow && conatinerStyles.containerNarrow}`}>
```

Then change the SCSS module to:

```scss
&.container-narrow {. <-- remove the space
```

The extra class will make it more specific than the general container component.

jasonlcrane: Thanks @kevee. I'll remove the redundant `narrow &&` part. 

On your second comment, take a look at the deploy preview (https://deploy-preview-474--upbeat-lovelace-3e9fff.netlify.com/about-project). I made `container-narrow` a child of the primary container so it stays on the page's  grid (keeping the content even with the page title). I may be missing a better solution, but applying the narrow style to the first container div centers it on the page, which is essentially the original change.

kevee: Ah gotcha, I'm fine with that.

kissane:
The only thing I see missing here, [looking at @beep's comp](https://github.com/COVID19Tracking/website/issues/452), is additional space between the header and the body content. Super psyched about this! Getting it just in time for our blog machine to pick it up.

jasonlcrane:
Thanks @kissane! I updated to add some additional space below the header (it's the same as above the footer, now). I see that @schwartzadev is planning to pick up the other changes from @beep's comp on #452, so I'm wondering if we should merge this and those changes together or let this one stand separately.

schwartzadev:
I'm flexible -- I'll tackle those changes tomorrow, if that makes any difference.

I would suggest keeping them separate, since it makes my git side of things a smidge easier, but I'm pretty agnostic.

Unless I hear otherwise before tomorrow, I'll plan on using a separate branch -- we can always merge the changes together before merging to master :+1:

beep:
If https://deploy-preview-474--upbeat-lovelace-3e9fff.netlify.com/about-project is the latest version, I think it’s a.) lookin’ _grand_ and b.) if @schwartzadev is willing to tackle further refinements separately, then we could get this shipped out the door!

schwartzadev:
@jasonlcrane I'm going to tap you to approve 0573f96738432f3ce065b2682b656e4e18d52af8 when you get a chance :+1:

jasonlcrane:
Nice improvement. Looks good to me 👍

Issue #479, opened 2020-04-10
Add toggle for "tests per capita" on sm multiples 


julialedur:
hey @vijaysharwar, 
could you add that toggle button for tests per capita on the sm multiples like we talked about yesterday? 

please follow style of the map buttons: 

<img width="286" alt="Screen Shot 2020-04-10 at 1 25 53 PM" src="https://user-images.githubusercontent.com/39416238/79009963-d1fdba80-7b2e-11ea-81c7-1a16790d9dd0.png">

we'll have "Total tests" and "Test per capita"

vijaysharwar:
Yep, I can work on this.

julialedur:
thank you! :)

vijaysharwar:
Anyone know why there are two different data files with population?

- [/src/data/visualization/state-populations.json](https://github.com/COVID19Tracking/website/blob/e3d26eb07d2c30e385f1a6c47ae2279aa8ccabcb/src/data/visualization/state-populations.json) used in [_MapContainer.js#L10](https://github.com/COVID19Tracking/website/blob/e3d26eb07d2c30e385f1a6c47ae2279aa8ccabcb/src/pages/dashboard/_MapContainer.js#L10)
- [/src/pages/dashboard/data/_state-populations.js](https://github.com/COVID19Tracking/website/blob/e3d26eb07d2c30e385f1a6c47ae2279aa8ccabcb/src/pages/dashboard/data/_state-populations.js) used in [_Map.js#L9](https://github.com/COVID19Tracking/website/blob/e3d26eb07d2c30e385f1a6c47ae2279aa8ccabcb/src/pages/dashboard/charts/_Map.js#L9)

Which is correct and/or should be used to normalize these state testing charts per capita? @goleary @jeremiak?

goleary:
the first one `/src/data/visualization/state-populations.json` is correct. If you could remove the other one and update the import that would be awesome!

vijaysharwar:
@goleary mind making that a separate cleanup task/issue, to keep scope of this concise?

Next issue is some of the territories don't have populations in the JS data set, namely:

- Guam
- Virgin Islands
- Marshall Islands
- American Samoa

For the time being, I can create a local hash map in [_StateCumulativeTestsContainer.js](https://github.com/COVID19Tracking/website/blob/dashboard-gatsby/src/pages/dashboard/_StateCumulativeTestsContainer.js) for these missing values, but it's a little brittle. Thoughts @goleary?

[...]

PR #494, opened 2020-04-10
Add toggle to switch between states' total tests versus per capita

vijaysharwar:
This adds a toggle that allows user to switch the state cumulative tests charts between total number of tests and per capita testing amount.

### Total tests:

tests-total

### Per capita:

tests-per-capita

### Notes:

- Existing state with population does not include the territories; added them as a [local hash in _StateCumulativeTestsContainer](https://github.com/vijaysharwar/website/blob/7e48dcd4db7e8249015bb3af02b809202b989212/src/pages/dashboard/_StateCumulativeTestsContainer.js#L72-L77)
- The data and state calling code (`MP`) are actually for "Northern Mariana Islands", not "Marshall Islands". [Renamed here](https://github.com/vijaysharwar/website/blob/7e48dcd4db7e8249015bb3af02b809202b989212/src/pages/dashboard/_utils.js#L40) as per conversation in https://github.com/COVID19Tracking/website/issues/479#issuecomment-612198044
- Changed `#map-style-button` to a [class named `.map-toggle`](https://github.com/COVID19Tracking/website/issues/479#issuecomment-612198044) for reusability. Side note: best practice usually dictates avoiding styling IDs for this precise reason 

Fixes COVID19Tracking/website#479

### DIFF
--- a/src/pages/dashboard/_StateCumulativeTestsContainer.js
+++ b/src/pages/dashboard/_StateCumulativeTestsContainer.js
@@ -1,5 +1,3 @@
-/* eslint-disable no-unused-vars */
-import { max } from 'd3-array'
 import { nest } from 'd3-collection'
 import { graphql, useStaticQuery } from 'gatsby'
 import React, { useMemo, useState } from 'react'
@@ -128,14 +126,6 @@ function groupAndSortStateDaily(query) {
     totals: sortGroupedData(grouped),
     perCapita: sortGroupedData(groupedPerCapita),
   }
-}
-
-function normalizeByStatePopulation(val, stateKey) {
-  if (!val) return 0
-  const statePop = statePopulations[stateKey]
-    ? statePopulations[stateKey]
-    : territoryPopulations[stateKey]
-  return val / statePop
 }
 
 export default function CumulativeTestsByStateContainer() {
--- a/src/pages/dashboard/_StateCumulativeTestsContainer.js
+++ b/src/pages/dashboard/_StateCumulativeTestsContainer.js
@@ -139,17 +139,15 @@ export default function CumulativeTestsByStateContainer() {
 
   const toggleChartData = () => setUseTestsPerCapita(u => !u)
 
-  const allData = useMemo(() => {
-    return groupAndSortStateDaily(query)
-  })
+  const allData = useMemo(() => groupAndSortStateDaily(query), [query])
 
   const data = useMemo(() => {
     return useTestsPerCapita ? allData.perCapita : allData.totals
-  })
+  }, [useTestsPerCapita])
 
   const maxStateTests = useMemo(() => {
     return data[0].values[0].totalTestResults
-  })
+  }, [useTestsPerCapita])
 
   return (
     <div>
--- a/src/pages/dashboard/_StateCumulativeTestsContainer.js
+++ b/src/pages/dashboard/_StateCumulativeTestsContainer.js
@@ -165,9 +165,7 @@ export default function CumulativeTestsByStateContainer() {
   })
 
   const maxStateTests = useMemo(() => {
-    return max(data, stateData => {
-      return max(stateData.values, d => calculateTotal(d))
-    })
+    return calculateTotal(data[0].values[0])
   })
 
   return (
--- a/src/pages/dashboard/_StateCumulativeTestsContainer.js
+++ b/src/pages/dashboard/_StateCumulativeTestsContainer.js
@@ -6,13 +6,7 @@ import cloneDeep from 'lodash/cloneDeep'
 import AreaChart from './charts/_AreaChart'
 import StatesWithPopulation from '../../data/visualization/state-populations.json'
 
-import {
-  calculateTotal,
-  getStateName,
-  parseDate,
-  totalColor,
-  positiveColor,
-} from './_utils'
+import { getStateName, parseDate, totalColor, positiveColor } from './_utils'
 
 import './dashboard.scss'
 
@@ -76,12 +70,7 @@ const territoryPopulations = {
 
 function sortGroupedData(groupedData) {
   return groupedData.sort((a, b) => {
-    const lastA = a.values[0]
-    const lastB = b.values[0]
-
-    const lastATotal = calculateTotal(lastA)
-    const lastBTotal = calculateTotal(lastB)
-    return lastBTotal - lastATotal
+    return b.values[0].totalTestResults - a.values[0].totalTestResults
   })
 }
 
@@ -115,6 +104,7 @@ function groupAndSortStateDaily(query) {
       // divide by population to determine per capita percentages
       clonedValue.positive /= population
       clonedValue.negative /= population
+      clonedValue.totalTestResults /= population
 
       return clonedValue
     })
@@ -138,6 +128,7 @@ export default function CumulativeTestsByStateContainer() {
             state
             positive
             negative
+            totalTestResults
           }
         }
       }
@@ -157,7 +148,7 @@ export default function CumulativeTestsByStateContainer() {
   })
 
   const maxStateTests = useMemo(() => {
-    return calculateTotal(data[0].values[0])
+    return data[0].values[0].totalTestResults
   })
 
   return (
@@ -239,7 +230,7 @@ export default function CumulativeTestsByStateContainer() {
             stateData.push({
               date,
               label: 'Total',
-              value: calculateTotal(d),
+              value: d.totalTestResults,
             })
           })
 
--- a/src/pages/dashboard/_StateCumulativeTestsContainer.js
+++ b/src/pages/dashboard/_StateCumulativeTestsContainer.js
@@ -148,7 +148,9 @@ export default function CumulativeTestsByStateContainer() {
 
   const toggleChartData = () => setUseTestsPerCapita(u => !u)
 
-  const allData = groupAndSortStateDaily(query)
+  const allData = useMemo(() => {
+    return groupAndSortStateDaily(query)
+  })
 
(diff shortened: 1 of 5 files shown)


src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -61,11 +64,35 @@ const stayAtHomeOrders = {
...
+  MP: 55144,
+  AS: 55641,
+}
+
+function sortGroupedData(groupedData) {
+  return groupedData.sort((a, b) => {
+    const lastA = a.values[0]
+    const lastB = b.values[0]
+
+    const lastATotal = calculateTotal(lastA)
```
goleary, line 84: the field `totalTestResults` contains `positive` + `negative` already so using this function is no longer necessary

src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -78,14 +105,37 @@ function groupAndSortStateDaily(query) {
...
+    return clonedData
   })
+
+  return {
+    totals: sortGroupedData(grouped),
+    perCapita: sortGroupedData(groupedPerCapita),
+  }
+}
+
+function normalizeByStatePopulation(val, stateKey) {
```
goleary, line 133: this doesn't appear to be called anywhere

src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -104,7 +154,15 @@ export default function CumulativeTestsByStateContainer() {
     }
   `)
 
-  const data = groupAndSortStateDaily(query)
+  const [useTestsPerCapita, setUseTestsPerCapita] = useState(false)
+
+  const toggleChartData = () => setUseTestsPerCapita(u => !u)
+
+  const allData = groupAndSortStateDaily(query)
```
goleary, line 161: this is performed on every render even though the value should never change. This one is a good candidate for `useMemo`.

Because the per cap toggle is the only thing that should trigger a rerender (only mutation to state in this component) the `useMemo` below doesn't accomplish anything. (not to mention you have to pass an array of dependencies for `useMemo` to work properly).

So I'd memoize `allData` and not `data` - although it doesn't hurt to memoize data too.

goleary:
image
do you also see that this area chart's size has been affected in this branch?

vijaysharwar:
@goleary RE your comment above https://github.com/COVID19Tracking/website/pull/494#issuecomment-612242783, which area chart? It looks the same as current `dashboard-gatsby`.

In https://github.com/COVID19Tracking/website/pull/449, I changed the `_AreaChart` to use `viewBox`. I suspect someone else was adding new charts without realizing that change, and the weird sizing is an effect of parallel PRs being merged.

src/pages/dashboard/_StateCumulativeTestsContainer.js
vijaysharwar: I am not seeing a `totalTestResults` field in this data. Inspecting the raw data:

Screen Shot 2020-04-10 at 6 34 55 PM

goleary: you'll have to modify the graphql query:
```
  const query = useStaticQuery(graphql`
    {
      allCovidStateDaily {
        edges {
          node {
            date
            state
            positive
            negative
            totalTestResults
          }
        }
      }
    }
  `)
```

goleary: also checkout http://localhost:8000/___graphql to make it easier to build your queries in the future.

src/pages/dashboard/_StateCumulativeTestsContainer.js
vijaysharwar: Good catch, fixed in 33658fd.

src/pages/dashboard/_StateCumulativeTestsContainer.js
vijaysharwar: Ah, yep. Memoized (and left `data` as is) in 14e451f.

src/pages/dashboard/_StateCumulativeTestsContainer.js
vijaysharwar: Ah, gotcha. Will update and remove the unnecessary calculations.

src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -104,12 +144,20 @@ export default function CumulativeTestsByStateContainer() {
...
   `)
 
-  const data = groupAndSortStateDaily(query)
+  const [useTestsPerCapita, setUseTestsPerCapita] = useState(false)
+
+  const toggleChartData = () => setUseTestsPerCapita(u => !u)
+
+  const allData = useMemo(() => {
+    return groupAndSortStateDaily(query)
+  })
```
goleary, line 153: unless you pass an array of dependencies to `useMemo` it will perform the calculation on every render.

so here you want `useMemo(()=>groupAndSortStateDaily(query),[query])`

https://reactjs.org/docs/hooks-reference.html#usememo

goleary: if only react was smart enough to just use the variables use from within the memoized function...

src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -104,12 +144,20 @@ export default function CumulativeTestsByStateContainer() {
...
+
+  const data = useMemo(() => {
+    return useTestsPerCapita ? allData.perCapita : allData.totals
+  })
 
   const maxStateTests = useMemo(() => {
-    return max(data, stateData => {
-      return max(stateData.values, d => calculateTotal(d))
-    })
+    return calculateTotal(data[0].values[0])
```
goleary, line 160: same thing here, this needs a dependency array

src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -104,12 +144,20 @@ export default function CumulativeTestsByStateContainer() {
...
+
+  const toggleChartData = () => setUseTestsPerCapita(u => !u)
+
+  const allData = useMemo(() => {
+    return groupAndSortStateDaily(query)
+  })
+
+  const data = useMemo(() => {
+    return useTestsPerCapita ? allData.perCapita : allData.totals
+  })
```
goleary, line 157: same here although `useMemo` here doesn't accomplish anything because `useTestPerCapita` changes on every rerender (it's the only thing that causes a rerender)

src/pages/dashboard/_StateCumulativeTestsContainer.js
vijaysharwar: Removed in e13d9ce.

goleary:
I see what you're saying, yeah that chart does look huge on the main branch. Don't worry about that in this PR then.

src/pages/dashboard/_StateCumulativeTestsContainer.js
vijaysharwar: @goleary check changes in 959f53a. I've memoized `allData` as suggested, and kept `useMemo` for the other two values (updating with `useTestsPerCapita`). Doing it this ensures that if other functionality is added, we're not resetting the `data` state for the charts unnecessarily. Thoughts?

julialedur:
this looks great, @vijaysharwar ! 🎉 
i have two thoughts: 

- move the toggle to the left, between title and legend
- make the legend change to "positive tests per capita" and "total tests per capita" when you click "Tests per capita" 

we should also make a tooltip for this chart. will open another issue for that.

src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -98,19 +128,26 @@ export default function CumulativeTestsByStateContainer() {
...
+    return useTestsPerCapita ? allData.perCapita : allData.totals
+  }, [useTestsPerCapita])
 
   const maxStateTests = useMemo(() => {
-    return max(data, stateData => {
-      return max(stateData.values, d => calculateTotal(d))
-    })
-  })
+    return data[0].values[0].totalTestResults
+  }, [useTestsPerCapita])
```
goleary, line 150: i would use `data` here instead of `useTestsPerCapita` since that's the variable that's used to produce this value.  That said the end result is no different so i'm not going to worry about this one.

goleary:
@julialedur i'm going to merge this to avoid conflicts and the follow ups your requested can be handled in additional PRs/commits.

goleary:
@julialedur I made the two bulletted adjustments:

image

fwiw I prefer the toggle on the right because I feel like it more effectively utilizes the space on the page but i'll leave it up to you.

julialedur:
thanks, @goleary! :) 
i agree that it uses the space better, but it also forces the user to read in an unnatural way, from right to left. 
i might be totally wrong about it (as UX is not my area of expertise) but we can discuss it with Sara once this is closer to ready!

vijaysharwar:
FWIW, I agree with @goleary that it would be best to  keep the legend and toggles in a row. Stacked how they are now is creating a lot of unnecessary dead space.

If the toggles are more important, they could go on the left, with the legend aligned right. This has additional UX sense, as it keeps the toggles in the same spot as the US map above. The legend is fine to move over, as it's more of an aside.

@julialedur let me know your thoughts, I can do a separate PR with this adjustment.

julialedur:
ok, let's to toggle on the left and legend on the right :)

vijaysharwar:
I'll work on that.

Issue #488, opened 2020-04-10
add note, source and bylines at the bottom 


julialedur:
please refer to [old dashboard](https://deploy-preview-136--upbeat-lovelace-3e9fff.netlify.com/dashboard/) and add those elements. 

Note should be updated to: 

**Notes:** We derive the total value by adding together the positive and negative value for each state. This is to account for differences in how states reporting pending tests; California has been inconsistent in its timing of reporting. We used faster updating sources until April 1, when we standardized on [California's new data dashboards](https://public.tableau.com/views/COVID-19PublicDashboard/Covid-19Public?:embed=y&:display_count=no&:showVizHome=no). This led to a drop in cases and deaths, as the state’s data lags some other sources.

Bylines will be updated with everyone's names later

AaronMullan:
PR incoming

PR #495, opened 2020-04-10
add note, source and bylines at the bottom #488

AaronMullan:


goleary:
i see the note & source but no byline

AaronMullan:
Whoops, thanks for catching that.  I was taking the instructions "Bylines will be updated with everyone's names later" to mean to not add the Byline for now.  But now that you point out that it is right there in the title of the Issue, it seems like the old byline should be there and it will be updated later.  Hold on, will update.

AaronMullan:
Updated the branch.

goleary:
looks good. Would you mind including a screenshot in the future so we can easily see what it looks like?

AaronMullan:
Will include screenshot.  Thanks for that.

PR #497, opened 2020-04-11
Focus finesser

AaronMullan:


kevee:
Hi, @AaronMullan - I think this should be a PR to one of the visualization branches, not `master`

PR #500, opened 2020-04-11
fix(visualiztion): brings back focus and finetunes it,re #429

AaronMullan:


src/pages/dashboard/map-container.scss
```diff
@@ -66,10 +60,13 @@ $death-color: #404856;
...
   font-weight: initial;
 }
+#toggle-container{
+  width: 100%;
+}
 
 .map-toggle {
   cursor: pointer;
-  width: 100%;
+  float: left;
```
goleary, line 69: I'd prefer to accomplish this with flexbox instead of floats

src/pages/dashboard/map-container.scss
```diff
@@ -66,10 +60,13 @@ $death-color: #404856;
   border: solid lightgray 2px;
   font-weight: initial;
 }
+#toggle-container{
```
vijaysharwar, line 63: I recommend using a class selector instead. Using IDs for CSS is considered bad practice. See https://stackoverflow.com/questions/8279132/why-shouldnt-i-use-id-selectors-in-css

We should maybe refactor all ID selectors throughout the SCSS. I've even noticed [nested IDs](https://github.com/COVID19Tracking/website/blob/dashboard-gatsby/src/pages/dashboard/map-container.scss#L101-L167) which _really_ overspecifies things. cc @goleary @beep @kevee

AaronMullan:
No problem using a class selector instead.

I can get the button to the left using justify-content: left; in the container div, but then the focus goes on the whole div again (see screenshot below).  

Any ideas?

Screen Shot 2020-04-11 at 10 09 01 AM

src/pages/dashboard/map-container.scss
kevee: Agreed!

AaronMullan:
Updated the PR, it now uses a class selector instead of a div id, and uses display: inline-block instead of float: left.

src/pages/dashboard/map-container.scss
vijaysharwar: Yeah, and also `float: left` breaks the layout of the other toggles, above the state charts further down the dashboard page.

Maybe try following how I did this with the other toggle. Keep a single parent `.map-toggle` as it was (with `width: 100%`). But add another class that sets `max-width` to avoid the container from stretching across its parent.

See:
- https://github.com/COVID19Tracking/website/blob/587c6622beb370976000386fd16f96f2e456e08c/src/pages/dashboard/dashboard.scss#L62-L66
- https://github.com/COVID19Tracking/website/blob/587c6622beb370976000386fd16f96f2e456e08c/src/pages/dashboard/_StateCumulativeTestsContainer.js#L164-L171

vijaysharwar:
@AaronMullan ah, `inline-block` seems to work. Ignore [my suggestion above](https://github.com/COVID19Tracking/website/pull/500#discussion_r407092798). In fact, you can probably remove that `.chart-tests-toggle` class now, and have both toggles markup/styles be the same. 🚀

src/pages/dashboard/map-container.scss
AaronMullan: Is that better practice than setting the element to display: inline-block?

vijaysharwar: @AaronMullan well, ideally we'd have styles that are completely reusable. I actually didn't like having to set `max-width` for the other toggles, but did so because I didn't want to mess with the existing class above :P

Another option might be using the flexbox container better, as I think @goleary is recommending. If you remove your parent `.toggle-container` entirely and set this on `.map-toggle` instead of `display: inline-block`:

```
  align-self: flex-start;
```

It should left align in the parent `#state-map` container. But then you'll need to apply similar styling to the parent container of the other toggles and state charts below, [this div](https://github.com/COVID19Tracking/website/blob/587c6622beb370976000386fd16f96f2e456e08c/src/pages/dashboard/_StateCumulativeTestsContainer.js#L153), something like:

```
  display: flex;
  flex-direction: column;
```

AaronMullan: I was able to make the map one work with align-self: flex-start;  I don't see any repercussions below but I could refactor them to work the same way.

vijaysharwar:
Nice, @AaronMullan. If someone can merge this, I can revise the toggles below to make 'em the same. I'm gonna change their position anyway as per convo here https://github.com/COVID19Tracking/website/pull/494#issuecomment-612486267.

@goleary can you re-review?

goleary:
@vijaysharwar done

PR #502, opened 2020-04-11
Fix tooltip tables styling

### LINKED ISSUE #477: Tooltip tables aren't styled anymore
Tooltip tables lost styling and alignment. What it looks like now: 

image (8)

What it should look like: 

Screen Shot 2020-04-02 at 1 32 22 AM

Slack chat:
re: the table regression, it could have to do with this: https://github.com/COVID19Tracking/website/issues/453 - there was some bleed-over between styles for the SummaryTable component used elsewhere in the site. Looks like that's been fixed now so some of the styles that were previously unintentionally inherited may need to be duplicated to the styles in src/pages/dashboard/charts/map.scss

nselikoff:
Fixes #477 

Screen Shot 2020-04-11 at 2 53 56 PM

One question before merging this: I noticed the per capita numbers in the tooltip table had lots of digits after the decimal point... I updated the formatNumber util to render them as integers. Let me know if this is not what is desired.

### DIFF
--- a/src/pages/dashboard/_utils.js
+++ b/src/pages/dashboard/_utils.js
@@ -3,7 +3,7 @@ import { format, formatPrefix } from 'd3-format'
 import { timeFormat, timeParse } from 'd3-time-format'
 
 export const formatDate = timeFormat('%b. %e')
-export const formatNumber = format(',')
+export const formatNumber = format(',.0f')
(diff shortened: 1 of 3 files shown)


julialedur:
great! 
that sounds good to me, @nselikoff ! 
especially cause you can't have half a test or half a death

goleary:
looks good!

Issue #501, opened 2020-04-11
Use classes instead of ID selectors


vijaysharwar:
Throughout the SCSS, there are examples of using ID selectors. This is a bad practice, as it creates overspecification and decreases modularity. We should replace all ID selectors with classes.

See conversation here: https://github.com/COVID19Tracking/website/pull/500#discussion_r407086177

Further reading: https://stackoverflow.com/questions/8279132/why-shouldnt-i-use-id-selectors-in-css

vijaysharwar:
@kevee @goleary @beep wondering further if the project should set SCSS style guidelines. Or consider some sort of convention like [OOCSS](http://oocss.org/) or [BEM](http://getbem.com/introduction/).

nselikoff:
@vijaysharwar looks like there's a first pass at style guidelines in Storybook; if you run `npm run storybook` then go here you'll see it: [http://localhost:6006/?path=/docs/styling--page](http://localhost:6006/?path=/docs/styling--page)

This could be expanded with info about classes instead of ID selectors.

Also, I'm seeing a few components that are using [CSS Modules](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/). It would be good to formalize whether we're going to use that pattern globally.

schwartzadev:
Style guidelines sound good to me -- I know that we recently moved from SCSS to SCSS modules, so there is certainly some scarring there.

As far as the classes/ID selectors, it looks like this has been resolved in 4088fa258ff74b4d899d7ab1cf52457ae3ac6fd3, so I'm going to close this issue.

With `grep -r "#.*:" components/ scss/` from the `src` directory, I'm not finding any matches in CSS, so I don't _think_ we're using this anymore on `master` or `dashboard-gatsby`.

nselikoff:
@schwartzadev there are IDs in use within the dashboard in the `dashboard-gatsby` branch (a lot of components haven't been extracted to `components/` yet)

PR #504, opened 2020-04-11
fix(visualization): Style map slider

### LINKED ISSUE #485: style map slider
let's tweak the slider style a bit and: 

- remove the gradient
- make slider fill #EDF1F2 and stroke #B7BCBD and thin 
- make circle fill #ffffff and stroke #000000 

it will look like this: 

Screen Shot 2020-04-10 at 2 01 47 PM

nselikoff:
Fixes #485 

image

Tested on Chrome, Safari and Firefox on OS X.


goleary:
I won't block on this for now but moving forward let's not use css id selectors as discussed in #501

PR #505, opened 2020-04-11
fix(visualization): Make sure outline of hovered state is on top

### LINKED ISSUE #352: change hover over style on both maps
right now our states turn gray when we hover over them. 
this is totally fine, except when we get to the choropleth, so suddenly the gray from the hover is getting in the way of the color encoding. 
so I suggest we only change the stroke around the state to black on mouseover. it's also easier to see which state you're looking at when thee shape is highlighted. Reuters did it well on [this](http://graphics.reuters.com/HEALTH-CORONAVIRUS-USA/0100B5K8423/index.html) story: 

Screen Shot 2020-04-02 at 1 18 38 AM

nselikoff:
Fixes #352 

Screen Shot 2020-04-11 at 5 05 39 PM

### DIFF
--- a/src/pages/dashboard/charts/_Map.js
+++ b/src/pages/dashboard/charts/_Map.js
@@ -227,22 +227,18 @@ const States = ({
       onMouseLeave={() => setHoveredState(null)}
     />
   ))
-  const hoveredStates = geoJson.features
-    .filter(d => hoveredState && hoveredState.state === d)
-    .map(d => (
-      <path
-        key={`path${d.properties.NAME}`}
-        d={path(d)}
-        className="hovered-states"
-        fill={getColorFromFeature(d)}
-        stroke="#000000"
-        strokeWidth="2px"
-      />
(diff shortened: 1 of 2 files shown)


goleary:
as you have it written a relatively expensive operation is computed on every render. It will iterate through the list of states and filter for the one which is selected.

We can optimize this by only displaying the hover element when a `hoverState` is set and further optimize by not iterating through the list or performing the filter since the geometery of the hoverState is already present within the `hoverState` variable.

Here what I suggest:

``` 
return (
    <>
      <g>{states}</g>
      {hoveredState && (
        <path
          d={path(hoveredState.state)}
          className="hovered-states"
          fill="transparent"
          stroke="#000000"
          strokeWidth="2px"
        />
      )}
    </>
  )
```

nselikoff:
Good call @goleary

goleary:
wonderful, thanks @nselikoff!

Issue #475, opened 2020-04-10
Area chart component doesn't respect it's width & height


goleary:
The AreaChart doesn't respect the height & width passed, it just seems to use them for proportions & relies on the container element to constrain it's size

AaronMullan:
PR #506

PR #506, opened 2020-04-11
fix(visualization): make area chart take width and height props #475

AaronMullan:


AaronMullan:
Screen Shot 2020-04-11 at 3 40 45 PM

src/pages/dashboard/charts/_AreaChart.js
```diff
@@ -60,7 +60,12 @@ const AreaChart = ({
   const strokeColor = '#b2bbbf'
 
   return (
-    <svg viewBox={`0 0 ${width} ${height}`} style={{ overflow: 'visible' }}>
+    <svg
+      width={width}
+      height={height}
```
vijaysharwar, line 65: This completely breaks the layout of the small area charts below. I had to switch this to using `viewBox` in order to scale the small area charts responsively across multiple breakpoints.

I think fix this instead by constraining the `AreaChart` in a container with the desired dimensions. It might even be preferable and feasible to put the two charts in a row together, like the ones in the "Differences between this data and the CDC data" section higher up in the dashboard.

vijaysharwar:
Please do not merge this PR without first [addressing comment above](https://github.com/COVID19Tracking/website/pull/506#pullrequestreview-391878827). I'm unable to request changes/prevent merge, as I'm not part of the org.

src/pages/dashboard/charts/_AreaChart.js
vijaysharwar: Screen Shot 2020-04-12 at 1 28 00 PM

PR #510, opened 2020-04-12
test(api): Add cumulative value testing for pos/neg daily values

gitjeff05:
This PR checks `api/v1/states/daily.json` to make sure that values for `positive` and `negative` are cumulative.

Running this test, I was able to see that there were some state errors in positive (RI) and negative (WI, DE, WY).

### DIFF
--- a/src/__tests__/api-test.js
+++ b/src/__tests__/api-test.js
@@ -1,0 +1,160 @@
+const fs = require('fs')
+
+/**
+ * Import the states daily json file
+ */
+const statesDaily = JSON.parse(
+  fs.readFileSync('./_data/v1/states/daily.json', 'utf8'),
+)
+
+/**
+ * List of states
+ */
+const uniqueStates = [...new Set(statesDaily.map(item => item.state))]
+
+/**
+ * These arrays capture the observed anomalies in cumulative testing results.
+ * Explanation:
+ * Since state websites serve as the AOR for cumulative data, their regressions
+ * result in errors in our cumulative positive/negative tests.
+ *
+ * We should verify these anomalies with QA/Data teams and then enter them here.
+ * These anomalies will be removed from the data in tests below so that these
+ * tests will pass. When new anomalies are found, they should be verified with
+ * QA/Data and only then should be entered into the arrays of allowed anomalies
+ * below.
+ */
+
+const allowedNegativeAnomalies = [
+  {
+    state: 'AK',
+    date: 20200413,
+    positive: 277,
+    negative: 7553,
+  },
+  {
+    state: 'AK',
+    date: 20200412,
+    positive: 272,
+    negative: 7766,
+  },
+  {
+    state: 'DE',
+    date: 20200410,
+    positive: 1326,
+    negative: 10415,
+  },
+  {
+    state: 'WI',
+    date: 20200330,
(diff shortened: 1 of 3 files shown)


gitjeff05:
Talked to some in the DE team about this and it sounds like cumulative values aren't always cumulative because our "source of truth" is the state website itself.

We could still use this -- perhaps with Jest's snapshot feature -- to capture the anomalies and just fail on new ones.

src/__tests__/api-test.js
```diff
@@ -0,0 +1,148 @@
...
+    positive: 1221,
+    negative: 15856,
+  },
+  {
+    state: 'WI',
+    date: 20200329,
+    positive: 1112,
+    negative: 16550,
+  },
+  {
```
gitjeff05, line 59: Example of an exception I had to add today to get tests to pass.

gitjeff05:
@webmasterkai and @kevee I'd like to get your input on this. Basically, these tests verify that all positive/negative values for _each_ state are cumulative. It adds ~100 tests. When any fail, we can see the state and values that caused the failure.

When there are anomalies, we would need to [add an exception](https://github.com/COVID19Tracking/website/pull/510/files#diff-8c46206e3a6f05e84796b87de4984da2R60) once it is verified -- by us or QA/DE -- or tests will fail.

It is hard to say how often exceptions would need to be added. There were about 4-5 this week but before that there were only 4 for all of March.

Will builds fail to deploy if `build:test` fails? I do not know enough about our build/CI pipeline.

src/__tests__/api-test.js
gitjeff05: Screen Shot 2020-04-17 at 5 55 21 AM

gitjeff05: When I see these, I just search the json to find the discrepency, then check the with QA/DE or check them myself by verifying the screenshots. If everything checks out, I add the exception so that tests pass.

kevee:
Hi, there, great work! Builds will fail if jest fails, so this might be something we want to capture in a separate test file - the test can fail and we can see that, but the website should still build successfully.

I am going to ping the data folks to see if they would like this level of QA work happening on this end.

webmasterkai:
There might be some overlap with what's happening here: https://github.com/COVID19Tracking/quality-control

gitjeff05:
Yeah, I was on the fence on whether or not this truly belongs here. Also it's not _really_ unit testing. But it _is_ where the csv/json for the API is built and deployed. Alas, I've reached out to the QA folks to see if there is overlap and if they need any help.

gitjeff05:
After talking to Joshua in QC, these tests are duplicated. However, I think there could be future benefits to having some API quality checks here. I'm going to suggest this:

1.) Keep these API quality checks here
2.) Move these tests to their own run script so they're non-blocking

gitjeff05:
@kevee Thoughts on keeping this around? It is catching anomalies (good) but may be a bit noisy and I don't know who would watch it. Other than QC, are there other efforts related to this front?

kevee:
I'm going to close this, the Data Entry team has taken on the QA part of the API side of things. Sorry that you did so much work on this, but I'm giving them your PR a sa model for where known issues are.

Issue #333, opened 2020-04-01
update stay-at-home orders on small multiples  


julialedur:
more states imposed stay-at-home orders, which are updated in this [spreadsheet](https://docs.google.com/spreadsheets/d/1mD_NhlJR1fM2Pv_pY8YixUrX2p2F8rAE0xPTtsTJOiM/edit#gid=0). we need to add those to the small multiples and also add a note at the bottom: 

Note: Only statewide stay-at-home orders were considered. The dates indicate when the order went into effect.

kevee:
@julialedur - was this fixed in #431?

kevee:
Ah the PR says it fixed this issue. I'm closing for now, feel free to reopen.

julialedur:
hey @kevee! 
this was fixed, but since then more states issued stay-at-home orders, so the graphic isn't updated.

PR #515, opened 2020-04-12
fix: Added stay at home date for MO, SC, TX

goldfarb:
Update stay at home order dates to include MO, SC, TX
[333](https://github.com/COVID19Tracking/website/issues/333)

Before
image

After 
image


julialedur:
looks great! thanks! 😄

julialedur:
oops closed it accidentally

PR #523, opened 2020-04-13
fix(no-js-nav): in-page links to states

### LINKED ISSUE #521: fix: no-js nav doesn't link to states
**Describe the bug**

To replicate:
1. Disable javascript
2. Go to the data index page
3. Click on one of the states in the state nav

The IDs in the no-js nav are currently of the form `#state-AK`, yet the actual item ids are of the form `#state-ak`, which seems to be breaking the in-page links.

The expected behavior would be to jump to the states lower down, which currently does not happen.

schwartzadev:
Fixed capitalization of links in the no JS nav, closes #521

### DIFF
--- a/src/components/pages/data/state-nav-no-script.js
+++ b/src/components/pages/data/state-nav-no-script.js
@@ -14,7 +14,7 @@ export default ({ stateList }) => (
       <ul>
         {stateList.map(state => (
           <li key={state.node.state}>
-            <Link to={`/data#state-${state.node.state}`}>
+            <Link to={`/data#state-${state.node.state.toLowerCase()}`}>
               {state.node.state}
             </Link>
           </li>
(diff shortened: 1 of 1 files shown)


src/components/pages/data/state-nav-no-script.js
```diff
@@ -14,7 +14,7 @@ export default ({ stateList }) => (
       <ul>
         {stateList.map(state => (
           <li key={state.node.state}>
-            <Link to={`/data#state-${state.node.state}`}>
+            <Link to={`/data#state-${state.node.state.toLowerCase()}`}>
```
kevee, line 17: Let's change these to regular `a` elements with an href so that there's no page load when you click on them:

```jsx
<a href={`#state-${state.node.state.toLowerCase()}`}>
```

schwartzadev: done in 5da01e54d50386663e05ea506b88b0c6e8f58989 😄

Issue #355, opened 2020-04-02
make maps bigger 


julialedur:
your maps are so beautiful and incredible and i think we should make them like 30% bigger on the page. 
let them shine! ❇️ 
plus it will also be easier to see the states on the east coast that are covered by bubbles

julialedur:
anyone working on this? @jeremiak @goleary @vijaysharwar @melbee

goldfarb:
I've resized the map, do we want to have a smaller version for mobile?

julialedur:
cool! looks great 🎉 
I think @goleary is working on the [mobile version](https://github.com/COVID19Tracking/website/issues/444).

PR #524, opened 2020-04-13
fix: resized map

goldfarb:
[355](https://github.com/COVID19Tracking/website/issues/355)

Resized the map

Before:
image

After:
image

### DIFF
--- a/src/pages/dashboard/charts/_Map.js
+++ b/src/pages/dashboard/charts/_Map.js
@@ -19,8 +19,8 @@ const margin = {
   right: 10,
   top: 10,
 }
-const height = 400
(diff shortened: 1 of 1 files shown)


src/pages/dashboard/charts/_Map.js
```diff
@@ -19,8 +19,8 @@ const margin = {
   right: 10,
   top: 10,
 }
-const height = 400
-const width = 700
+const height = 520
```
jeremiak, line 22: This totally works, thanks! I wonder if we should maybe do something like get the width of the document and then make the width related to that? Let's merge this for now and reconsider that later

PR #511, opened 2020-04-12
Move chart legend to right of toggle on large screens

vijaysharwar:
Updating the small area charts toggle and legend to make better use of available space on larger screens. See conversation in https://github.com/COVID19Tracking/website/pull/494#issuecomment-612313777.

### Notes:
- Uses updated toggle styles from https://github.com/COVID19Tracking/website/pull/500
- Moves `.map-toggle` to `.dashboard-toggle` and into `dashboard.scss`, since this
  component is shared between the map and small area charts

### Small screens (< 960px):

Screen Shot 2020-04-12 at 1 13 15 PM

### Large screens (960px and above):

Screen Shot 2020-04-12 at 1 13 57 PM

cc @goleary @AaronMullan @julialedur

PR #526, opened 2020-04-13
fix(dashboard): improve map display on mobile and use global colors

goleary:
Bubble Map:
image

Choropleth:
image


goleary:
i believe some hard coded colors live on elsewhere in the dashboard page that'll we'll need to DRY up.

src/pages/dashboard/map-container.scss
```diff
@@ -32,41 +33,79 @@ $death-color: #404856;
...
+    &:nth-child(2) {
+      border-radius: 8px 0 0 8px;
+      border-right-color: black !important;
+    }
+    &:nth-child(1) {
+      border-radius: 0 8px 8px 0;
+    }
+    &.active {
+      @extend %map-toggle-active;
+    }
```
vijaysharwar, line 78: @goleary I moved/renamed these styles to `dashboard.scss` since they're used elsewhere on the page, not just for the map. Might need to update this feature branch with what's already merged to `dashboard-gatsby`.

See https://github.com/COVID19Tracking/website/pull/511/files#diff-d99b5a549ac78593a54c2878d296386c.

src/pages/dashboard/charts/_Map.js
```diff
@@ -144,26 +128,37 @@ export default function Map({
     [maxValue],
   )
 
+  const isMobile = window.screen.availWidth < viewportSm
```
vijaysharwar, line 131: Could be an enhancement, but it might be good to throw a window resize listener on this (with a debounce) and resent `isMobile`. Not a huge use case for laptop/desktop users, but could/should change if mobile user changes orientation.

src/pages/dashboard/charts/_Map.js
```diff
@@ -144,26 +128,37 @@ export default function Map({
...
           />
         ) : (
-          <BubbleLegend data={data} r={r} maxValue={maxValue} />
+          <BubbleLegend
+            data={data}
+            r={r}
+            maxValue={maxValue}
+            height={isMobile ? 100 : 150}
+            width={isMobile ? 100 : 150}
+          />
```
vijaysharwar, line 156: I'm checking things in Chrome dev tools with iPhone X dimensions, and this element is falling offscreen to the right:

Screen Shot 2020-04-13 at 5 08 27 PM

The legend seems less important for mobile. Wondering if we can:

- move it above or below the toggle, all elements stacked vertically
- move it below the map
- hide it entirely on small screens

vijaysharwar:
@goleary this PR resizes the other charts, might help with checking overall responsiveness: https://github.com/COVID19Tracking/website/pull/527

src/pages/dashboard/charts/_Map.js
goleary: interesting, on my machine with iphone x dimensions this is what I see:
image
I'm going to wait for the chart size changes from #527 to see how they play together before readdressing.

src/pages/dashboard/charts/_Map.js
goleary: added #528

Issue #503, opened 2020-04-11
make CDC charts smaller and centered


julialedur:
<img width="1270" alt="Screen Shot 2020-04-11 at 2 56 51 PM" src="https://user-images.githubusercontent.com/39416238/79052459-b238c600-7c04-11ea-8f28-b208fb00601a.png">

- let's make these first two charts 40% smaller on the page
- not so far apart 
- and centered on the page

goldfarb:
- I'm going to work on this -

Issue #484, opened 2020-04-10
grids and labels colors 


julialedur:
let's tweak the style of our labels and grids: 

- labels should be #828688

- grids should be #B7BCBD and 0.7

vijaysharwar:
I'm working on this one.

julialedur:
bubble legend style moved to [this issue](https://github.com/COVID19Tracking/website/issues/593).

PR #527, opened 2020-04-13
Update barchart size 503

goldfarb:
[503](https://github.com/COVID19Tracking/website/issues/503)

- Made the CDC bar charts smaller
- Made the bar chart text smaller 
- Added an `align` property to the bar chart

Before:
image

After:
image

After, for the Total cumulative deaths chart
image

### DIFF
--- a/src/pages/dashboard/charts/_BarChart.js
+++ b/src/pages/dashboard/charts/_BarChart.js
@@ -16,6 +16,7 @@ const BarChart = ({
   marginTop,
   xTicks,
   width,
+  align,
   yMax,
   yTicks,
 }) => {
@@ -38,7 +39,7 @@ const BarChart = ({
     Math.floor(xScaleDomain.length / xTicks),
   )
   return (
-    <div>
+    <div align={align}>
       <svg width={width} height={height} viewBox={`0 0 ${width} ${height}`}>
         <g
           className="axis-group"
--- a/src/pages/dashboard/charts/_BarChart.js
+++ b/src/pages/dashboard/charts/_BarChart.js
@@ -47,7 +47,11 @@ const BarChart = ({
           <g className="chart-grid">
             {yScale.ticks(yTicks).map(tick => (
               <g key={tick}>
-                <text y={yScale(tick) + 6} x={`${tick}`.length * -11}>
+                <text
+                  y={yScale(tick) + 6}
+                  x={`${tick}`.length * -11}
+                  fontSize="smaller"
+                >
                   {formatNumber(tick)}
                 </text>
                 <line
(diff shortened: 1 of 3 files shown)


goldfarb:
I don't love what the x-axis label is doing 

image

& will try to get that sorted out later today / if someone else could take a look at it

goldfarb:
I tried to emulate the code style for the changes I made, but may not have done it with adding the [alignment](https://github.com/COVID19Tracking/website/pull/527/commits/0a0b5c1f1d74ad21bdb695d0a7e15d44287e3ee5) or [the chart margin](https://github.com/COVID19Tracking/website/pull/527/commits/6c712bbf8598e5dcd071b7c9765407e65f4c2c19), so I'd like a double check on the code as well as the outcome, please.

julialedur:
thanks, @goldfarb! 🎉 

I'd make it like 5% bigger, maybe that fixes the labels problem. 
let's do a few other things: 

- leave only 0 and 50,000 labels and grids for the first chart (delete the rest) 
- on the x-axis, can we label Feb Mar and Apr on the first day of the month? some ticks would be helpful too so we can see exactly which bar is labeled 
- I have an [issue for grids and labels styles](https://github.com/COVID19Tracking/website/issues/484), but would be great if you could fix those for these charts while you're at it (font-size of labels will probably be smaller for these) 

let me know if this makes sense!

goldfarb:
- Readjusted the chart size - made the bar charts 5% larger

- For the CDC chart, the removed the labels and grids > 50,000



Before
image

After
image

And made adjustments so the cumulative deaths chart is not effected
image

src/pages/dashboard/charts/_BarChart.js
```diff
@@ -37,21 +40,27 @@ const BarChart = ({
     xScaleDomain.length,
     Math.floor(xScaleDomain.length / xTicks),
   )
+  const textColor = { textColor: colors.text }
   return (
-    <div>
+    <div align={align}>
```
jeremiak, line 45: What does the `align` attribute do in this case on the `div`?

julialedur:
if we have time, can we change the x-axis labels to Feb Mar Apr (on the first day of each month)?

src/pages/dashboard/charts/_BarChart.js
goldfarb: the `align` allows for left / center / right alignment within the surrounding `div`, so that the chart on the left (CDC data) can be right aligned and the right hand chart (tests administered) can be left aligned, so together they are in the center of the page

image

jeremiak: Woah, maybe I'm missing something but how does it do that??

PR #535, opened 2020-04-14
fix: text-heavy page design

schwartzadev:
This PR resolves the design fixes for text-based pages, per [Ethan's designs](https://www.figma.com/file/fKUKsYMNw3NLSyIEOI1HKc/v2-refresh) in #452 

Here are some screenshots of this design:
image
image
image
image

@beep I would _love_ for you to review this :+1: 

Edit: since the scss changes are at the container level, they hit other pages as well:
image

I would suggest making a scss class for text-heavy pages and applying these styles there. @kevee what do you think?

### DIFF
--- a/src/components/common/container.module.scss
+++ b/src/components/common/container.module.scss
@@ -27,6 +27,7 @@
     }
     h2 {
       font-size: 30px;
+      font-size: 1.875rem;
       line-height: 33px;
       margin: 2rem 0 1rem;
       padding-top: 1em;
--- a/src/components/common/container.module.scss
+++ b/src/components/common/container.module.scss
@@ -29,8 +29,7 @@
       font-size: 30px;
       font-size: 1.875rem;
       line-height: 37.5px;
-      margin: 2rem 0 0.5rem;
-      padding-top: 1em;
+      margin: 4rem 0 0.5rem;
     }
     h3 {
       margin-bottom: 1em;
--- a/src/components/common/container.module.scss
+++ b/src/components/common/container.module.scss
@@ -26,6 +26,9 @@
     margin: 2rem 0 1rem;
     padding-top: 1em;
   }
+  h3 {
+    margin-bottom: 1em;
+  }
   li {
     margin-bottom: 0.75rem;
   }
--- a/src/components/common/container.module.scss
+++ b/src/components/common/container.module.scss
@@ -19,6 +19,7 @@
     max-width: 42rem;
   }
   &.text-heavy {
+    margin: 5em auto;
     @media (min-width: $viewport-lg) {
       font-size: 22px;
       line-height: 1.5;
--- a/src/components/common/container.module.scss
+++ b/src/components/common/container.module.scss
@@ -41,7 +41,7 @@
     p,
     ul,
     ol {
-      margin-bottom: 2em;
+      margin-bottom: 1.6em;
     }
   }
 }
--- a/src/components/common/container.module.scss
+++ b/src/components/common/container.module.scss
@@ -29,7 +29,7 @@
       font-size: 30px;
       font-size: 1.875rem;
       line-height: 37.5px;
-      margin: 2rem 0 1rem;
+      margin: 2rem 0 0.5rem;
       padding-top: 1em;
     }
     h3 {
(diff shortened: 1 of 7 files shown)


schwartzadev:
So these latest commits add a param to identify a `Layout` as text-heavy, and then applies the appropriate styles. The affected pages are:
* src/pages/about-project/in-the-press.js
* src/pages/blog.js
* src/templates/blog-post.js
* src/templates/content.js

This fixes the problem with these new styles applying to data pages, etc.

src/components/common/container.module.scss
```diff
@@ -18,4 +18,28 @@
   & .container-narrow {
     max-width: 42rem;
   }
+  &.text-heavy {
+    margin: 5em auto;
+    @media (min-width: $viewport-lg) {
+      font-size: 22px;
```
beep, line 24: How would you feel about converting this to ems, @schwartzadev? Or if you prefer, rems with a px fallback:

```
font-size: 22px;
font-size: 1.375rem;
```

Totally defer to you on what’s most feasible here! But a relative font size’ll be a bit more accessible.

src/components/common/container.module.scss
```diff
@@ -18,4 +18,28 @@
...
+      line-height: 1.5;
+    }
+    h2 {
+      font-size: 30px;
```
beep, line 28: Same question [as above](https://github.com/COVID19Tracking/website/pull/535/files#diff-568a46f3cee58336c4574d08e1c143f5R24) regarding relative font-sizing!

src/components/common/container.module.scss
```diff
@@ -18,4 +18,28 @@
...
+      line-height: 33px;
```
beep, line 29: I think my 1.1 leading was a little too aggressive; could you bump this up to 1.25?

src/components/common/container.module.scss
```diff
@@ -18,4 +18,28 @@
...
+    h3 {
+      margin-bottom: 1em;
+    }
+    li {
+      margin-bottom: 0.75rem;
+    }
+    p,
+    ul,
+    ol {
+      margin-bottom: 2em;
```
beep, line 42: I think this could be tightened up a bit to `1.6em`.

src/components/common/container.module.scss
```diff
@@ -18,4 +18,28 @@
...
+  &.text-heavy {
+    margin: 5em auto;
+    @media (min-width: $viewport-lg) {
+      font-size: 22px;
+      line-height: 1.5;
+    }
+    h2 {
+      font-size: 30px;
+      line-height: 33px;
+      margin: 2rem 0 1rem;
```
beep, line 30: This bottom margin feels a _touch_ too big to me; maybe `0.5rem`?

src/components/common/container.module.scss
```diff
@@ -18,4 +18,28 @@
...
+      padding-top: 1em;
```
beep, line 31: Two small questions!

1. Do we need the top padding _and_ the top margin? Is this is here for margin collapsing reasons, or could we consolidate the top-edge spacing to the `margin` property?
2. I think this _might_ have introduced a slight layout jog with the permalink on headers? Here’s a screenshot:

Screen Shot 2020-04-14 at 2 54 30 PM

src/components/common/container.module.scss
schwartzadev: Great call!

I kept the px fallback and switched to rem in 83a50c606df97a02a50c56da914aca24f5ef2ce5

src/components/common/container.module.scss
schwartzadev: fixed in bdfd3485057bbe93c14f4b1ec4255617e5b139d1

src/components/common/container.module.scss
schwartzadev: 1. Nope -- I collapsed the spacing into `margin` in d274846cd0e1d02c30ee335e167211f5b8d5578c. (Before, we had a vertical space of `69.75px`. Now, we have a vertical space of `72px` using the `4em` margin.)
2. Yes. Fixed!
image

beep:
oh wait i can probably do the thing, i can hit the merge thing, hi

PR #544, opened 2020-04-14
Dashboard normalize choro

### LINKED ISSUE #534: "per million people on map" is misleading
@julialedur 
This makes it look like the US has 23k deaths per million people.

image

The number presented here is absolute.  Options:
- make the number per capita
- move the text elsewhere (maybe that's accounted for in what you and @jeremiak have been discussing)

goleary:
closes #534


goleary:
accidentally branched of #540

Issue #533, opened 2020-04-14
Area chart is way too big


jeremiak:
The area chart we use under the heading of "Positive tests and total tests in the US" is far too big for the page, making it hard to read. Let's make the width consistent with our other charts.

As you can see in the screenshot, I can't get the whole chart in a single shot.
<img width="1401" alt="Screen Shot 2020-04-14 at 1 20 51 PM" src="https://user-images.githubusercontent.com/780941/79254568-eb667580-7e52-11ea-94a5-b32640056d35.png">

goldfarb:
-- I'll give this a shot --

Issue #539, opened 2020-04-14
put death charts side by side 


julialedur:
we need to layout the death charts side by side (US cumulative on the left)

also, let's change US cumulative chart to an **area** chart and #4C5559

PR #547, opened 2020-04-14
Update chart size 533

goldfarb:
[533](https://github.com/COVID19Tracking/website/issues/533)

Adjusted the size for the Positive tests and total tests in the US chart

image


julialedur:
awesome! can we center it?

goleary:
@julialedur aren't we putting them side by side anyway? or am I misunderstanding #539

julialedur:
I was talking about  **Positive tests and total tests in the US chart** up there 

yes, we should put the 2 **death charts** side by side

PR #549, opened 2020-04-14
fix(dashboard): cdc data now starts on same date as other data

AaronMullan:


src/pages/dashboard/_CDCComparisonContainer.js
```diff
@@ -42,6 +42,7 @@ export default function CDCComparisonContainer() {
       }
     })
     .sort(sortChronologically)
+    .filter(element => element.date > new Date(2020, 2, 3))
```
vijaysharwar, line 45: Hmm, I think this is somewhat fragile. Hardcoding the date assumes that earlier Covid Tracking data doesn't become available. I also noticed that the current data set for CDC ends before the Covid Tracking data... which creates mismatches on the right end of the time scale.

I'm working on a solution that normalizes things. It would even work if older Covid Tracking data comes to light.

AaronMullan:
@vijaysharwar the approach you described in the related issue sounds like a better solution.  I'll leave this here for now in case somebody wants the quick/dirty version for the moment.

vijaysharwar:
It might be beneficial to do both. The number of tests before early Feb are so low that it's practically meaningless; they aren't visible in the charts at all. I think it comes down to what story we're trying to tell. Might need @julialedur's input.

julialedur:
hey all, i think the charts should have the same date range, yes! it's okay if numbers are low in the beginning

vijaysharwar:
The underlying issue (mismatched x-axes) has been resolved via https://github.com/COVID19Tracking/website/pull/552. I think we can close this without merge, as @julialedur indicated it's okay that the charts don't show significant bars/numbers before February.

PR #553, opened 2020-04-14
fix(dashboard): focus is now dotted, with color imported from colors.…

AaronMullan:
…scss


AaronMullan:
I imported the color for the :focus because I think the site is moving in that direction.  Hopefully that's the right move, otherwise it can just be written in as a hex value (#D0D7DB)

PR #556, opened 2020-04-15
fix(dashboard): Makes the div containing the two death charts space better

AaronMullan:
I think this looks better on the page.
Screen Shot 2020-04-14 at 5 56 35 PM


src/pages/dashboard/index.js
```diff
@@ -130,7 +130,7 @@ const DashboardPage = () => {
         <a href="https://twitter.com/jburnmurdoch">John Burn-Murdoch</a>, a data
         journalist at the <em>Financial Times</em>, pointed out.
       </p>
-      <div style={{ display: 'flex' }}>
+      <div style={{ display: 'flex', justifyContent: 'space-around' }}>
```
goleary, line 133: will be good to get all css into modules but there's more of that to do elsewhere so i'll let it slide for now.

PR #558, opened 2020-04-15
fix(dashboard): stops text from changing color when focused on

AaronMullan:


src/pages/dashboard/dashboard.scss
```diff
@@ -251,5 +251,5 @@
 
 :focus {
   outline: 2px dotted;
-  color: $color-slate-9;
+  outline-color: $color-slate-9;
 }
```
vijaysharwar, line 255: I'm unsure about this block more generally. Without a class or any other scoping, this will affect *all* elements on the page. Is that what we want? If so, it probably belongs in a global stylesheet, not in this partial `dashboard.scss` file.

goleary: this is a good point. The effects of this have bled to the rest of the site:
image

AaronMullan:
Does anyone have thoughts on how to scope?  I'm confused because some of the 'buttons' aren't html buttons (like the bubble map/cloropleth map slider).

AaronMullan:
I guess I could make a class and just apply that to any element we want ...and looks like you've already done that...thanks @jeremiak .

jeremiak:
Np, thanks for debugging the issue @AaronMullan! I'm going to close this one now

PR #559, opened 2020-04-15
Limit changes on :focus

jeremiak:
An alternative to #558


goleary:
This only addresses the higlighting on the toggles
There are other places we see this issue:
play/stop button:
image

expand buttons (state death chart & small multiples)
image

Issue #486, opened 2020-04-10
add titles and subtitles to graphics and section headers


julialedur:
add titles and subtitles (when applicable) to all graphics. 
 
please use the titles on these mockups (use same styles): 

![mockup-part1](https://user-images.githubusercontent.com/39416238/79012608-ad0c4600-7b34-11ea-92a8-57aa50946af3.jpg)

![mockup-part2](https://user-images.githubusercontent.com/39416238/79012621-b09fcd00-7b34-11ea-9367-d0306899d281.jpg)

![mockup-part3](https://user-images.githubusercontent.com/39416238/79012626-b4335400-7b34-11ea-961d-2017d8024289.jpg)

- please also add section headers ("Data quality varies across states", "Comparing states", and "The gray area of death counts")

PR #560, opened 2020-04-15
feat(visualization): Added new titles

sakuraaii:
This pull request adds titles to graphics as suggested in #486.

### DIFF
--- a/src/pages/dashboard/_CDCComparisonContainer.js
+++ b/src/pages/dashboard/_CDCComparisonContainer.js
@@ -66,8 +66,13 @@ export default function CDCComparisonContainer() {
           sample.
         </p>
       </div>
+      <h3>DAILY NEW TESTS IN THE U.S.</h3>
       <div style={{ display: 'flex' }}>
         <div style={{ flexGrow: 1, width: '50%' }}>
+          <div>
+            <h4 style={{ marginBottom: '0.02rem' }}>C.D.C.</h4>
+            <p>Specimens tested</p>
+          </div>
           <BarChart
             data={cdcData}
             fill="#585BC1"
@@ -82,6 +87,10 @@ export default function CDCComparisonContainer() {
           />
         </div>
         <div style={{ flexGrow: 1, width: '50%' }}>
+          <div>
+            <h4 style={{ marginBottom: '0.02rem' }}>Covid tracking project</h4>
+            <p>People tested</p>
+          </div>
           <BarChart
             data={data}
             fill="#585BC1"
--- a/src/pages/dashboard/_MapContainer.js
+++ b/src/pages/dashboard/_MapContainer.js
@@ -136,6 +136,7 @@ const MapContainer = () => {
 
   return (
     <div className="state-map">
+      <h3>CUMULATIVE TESTS IN THE U.S.</h3>
       <div className="map-dek">
         <h2>{formatDate(parseDate(currentDate))}</h2>
         {useChoropleth ? (


src/pages/dashboard/_CDCComparisonContainer.js
```diff
@@ -66,8 +66,13 @@ export default function CDCComparisonContainer() {
           sample.
         </p>
       </div>
+      <h3>DAILY NEW TESTS IN THE U.S.</h3>
```
jeremiak, line 69: Let's make this normal text case and use CSS to transform it. Also, I know its weird but we're using "US" instead of "U.S." for consistency

src/pages/dashboard/_MapContainer.js
```diff
@@ -136,6 +136,7 @@ const MapContainer = () => {
 
   return (
     <div className="state-map">
+      <h3>CUMULATIVE TESTS IN THE U.S.</h3>
```
vijaysharwar, line 139: We have a different `h3` added here upstream at @julialedur's request. I think this is now incorrect.

https://github.com/COVID19Tracking/website/blob/a3c01a98245792e168daffc4c5ded5ac38b554a0/src/pages/dashboard/_MapContainer.js#L126-L128

src/pages/dashboard/_CDCComparisonContainer.js
vijaysharwar: This is already added upstream with more recent styling.

https://github.com/COVID19Tracking/website/blob/a3c01a98245792e168daffc4c5ded5ac38b554a0/src/pages/dashboard/_CDCComparisonContainer.js#L129-L131

src/pages/dashboard/_CDCComparisonContainer.js
```diff
@@ -66,8 +66,13 @@ export default function CDCComparisonContainer() {
...
       <div style={{ display: 'flex' }}>
         <div style={{ flexGrow: 1, width: '50%' }}>
+          <div>
+            <h4 style={{ marginBottom: '0.02rem' }}>C.D.C.</h4>
```
vijaysharwar, line 73: Let's avoid inline styles in favor of classes (preferably in the new `dashboard.module.scss`).

src/pages/dashboard/_CDCComparisonContainer.js
```diff
@@ -82,6 +87,10 @@ export default function CDCComparisonContainer() {
           />
         </div>
         <div style={{ flexGrow: 1, width: '50%' }}>
+          <div>
+            <h4 style={{ marginBottom: '0.02rem' }}>Covid tracking project</h4>
```
vijaysharwar, line 91: Avoid inline styles, and also we now have logic upstream that switches this title depending on the viewport to avoid line breaks.

https://github.com/COVID19Tracking/website/blob/a3c01a98245792e168daffc4c5ded5ac38b554a0/src/pages/dashboard/_CDCComparisonContainer.js#L151

src/pages/dashboard/_CDCComparisonContainer.js
```diff
@@ -66,8 +66,13 @@ export default function CDCComparisonContainer() {
           sample.
         </p>
       </div>
+      <h3>DAILY NEW TESTS IN THE U.S.</h3>
       <div style={{ display: 'flex' }}>
         <div style={{ flexGrow: 1, width: '50%' }}>
+          <div>
+            <h4 style={{ marginBottom: '0.02rem' }}>C.D.C.</h4>
+            <p>Specimens tested</p>
```
vijaysharwar, line 74: An `<h5>` would be more semantic, as this is a subheading to the heading above.

kevee:
@sakuraaii - Closing this PR as it should be a request against master.

PR #583, opened 2020-04-15
Post-blog launch updates

schwartzadev:
Cleans up the blog page

### DIFF
--- a/src/pages/blog.js
+++ b/src/pages/blog.js
@@ -30,6 +30,7 @@ export const query = graphql`
           slug
           author {
             name
+            twitterLink
           }
           publishDate(formatString: "MMMM D, YYYY")
           lede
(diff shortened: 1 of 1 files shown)


src/pages/blog.js
```diff
@@ -4,18 +4,16 @@ import Layout from '../components/layout'
...
           <Link to={`/blog/${node.slug}`}>{node.title}</Link>
-        </h3>
+        </h2>
         <Byline author={node.author} date={node.publishDate} />
         <div
           dangerouslySetInnerHTML={{
-            __html:
-              node.childContentfulBlogPostTeaserTextNode.childMarkdownRemark
-                .html,
+            __html: node.lede,
```
kevee, line 16: Since lede is regular text, let's get rid of this div and just use:

`<p>{node.lede}</p>`

schwartzadev: Good call, yep. I'm adding the same `lede` class as we have in the individual post pages, too.

Issue #563, opened 2020-04-15
changes to US death charts 


julialedur:
<img width="463" alt="Screen Shot 2020-04-15 at 11 09 02 AM" src="https://user-images.githubusercontent.com/39416238/79354143-03e59700-7f0a-11ea-8108-37e94c2c4f4f.png">

- bars on this chart should be #4C5559 
- let's change it to **daily new deaths** instead of cumulative

goldfarb:
- I'll work on this -

PR #586, opened 2020-04-15
fix(dashboard): update US death chart color 563

goldfarb:
[563](https://github.com/COVID19Tracking/website/issues/563)

Before:
image

After:
image

Changed the color
Changed the data source to the daily increase, rather than the cumulative death count
Changed the title
Moved the count label 1 to the left, to keep the number from overlapping the grid line (and checked that it didn't create problems for the other bar charts)


src/pages/dashboard/_UsCumulativeDeathsContainer.js
```diff
@@ -36,11 +36,11 @@ export default function UsAreaChartContainer() {
   return (
     <section>
       <div>
-        <h4>Total cumulative deaths by day in the US</h4>
+        <h4>Daily deaths in the US</h4>
         <div>
           <BarChart
             data={data}
-            fill="#585BC1"
+            fill="#4C5559"
```
goleary, line 43: we should use a color imported from the global color file `src/scss/colors.scss`

src/pages/dashboard/_UsCumulativeDeathsContainer.js
```diff
@@ -1,15 +1,15 @@
 import React from 'react'
```
goleary, line 1: we should change the name of this file

Issue #569, opened 2020-04-15
subhead and headers styles 


julialedur:
**Subhead** 
- use these styles for subhead: 
<img width="167" alt="Screen Shot 2020-04-15 at 11 46 33 AM" src="https://user-images.githubusercontent.com/39416238/79358205-ecf57380-7f0e-11ea-9ede-ca999b6b9a6e.png">

- add margin-top: 50px; to first paragraph after subhead 
<img width="1069" alt="Screen Shot 2020-04-15 at 11 46 07 AM" src="https://user-images.githubusercontent.com/39416238/79358115-d6e7b300-7f0e-11ea-9a53-6c964161ee9f.png">

**Graphic titles** 
apply text-transform: uppercase; and font-size: 16px; to all graphic titles like "TOTAL DEATHS BY STATES". 
for "CDC" and "THE COVID TRACKING PROJECT" subheaders in the first chart do font-size: 13px;

**Guidelines titles**
For all tips, like "Choose colors carefully." do font-size: 13px;

**Notes** 
apply font-size: 14px; and color: #676B6D; to all notes, *s, and to notes, source and byline at the bottom of the page

goldfarb:
-- I'll work on this --

PR #601, opened 2020-04-16
Update headers style 569

goldfarb:
[569](https://github.com/COVID19Tracking/website/issues/569)

Updated subhead style

image

Updated chart head and subhead style

image

Updated notes, and guideline titles

image


julialedur:
awesome! 🎉 can we make the chart titles bold? 

also, this was my bad, but the guideline titles need to be a bit bigger than the body-text

goldfarb:
Updated guideline title size

image

Updated chart title to bold 

image

jeremiak:
I'm confused why this has been pending a check for a long time and then when I go see what "checks" are being run Github can't find any for this commit. So I'm going to merge this since its a straightforward change.

Issue #577, opened 2020-04-15
remove periods from months on the page


julialedur:
we need to remove the period in all months to match our [style guide](https://docs.google.com/document/u/5/d/1fi_v-Bs5R3a8e3mqFe9Lz7C_LYAF49FI3KMsS1h-nqw/edit#). 
so instead of "Mar. 27", let's do "Mar 27"

(don't forget the map tooltip!)

AaronMullan:
I will do this at 2:30 Pacific today if it's not done before.

AaronMullan:
Oh wait I think it was super easy.  PR #612

PR #612, opened 2020-04-16
fix(visualization):remove periods from months on the page #577

AaronMullan:


AaronMullan:
Screen Shot 2020-04-16 at 1 01 51 PM

PR #618, opened 2020-04-16
fix(charts): 573/chart headers alignment

### LINKED ISSUE #571: Streamline dashboard chart styles
Each of the chart sections on the dashboard use different CSS classes and styles. The location of styles is also not consistent. For example, some charts use classes set in [dashboard.scss](https://github.com/COVID19Tracking/website/blob/d4d6def2a9754bdff0ef1a4e44a3998834dd94de/src/pages/dashboard/dashboard.scss) while others have their own partial SCSS files, like [state-cumulative-deaths-container.scss](https://github.com/COVID19Tracking/website/blob/d4d6def2a9754bdff0ef1a4e44a3998834dd94de/src/pages/dashboard/state-cumulative-deaths-container.scss), or [charts/area-chart.scss](https://github.com/COVID19Tracking/website/blob/dashboard-gatsby/src/pages/dashboard/charts/area-chart.scss).

For all charts:

- Daily new tests in the US
- Positive tests and total tests in the US
- Total deaths in the US
- Total Deaths By States
- Cumulative tests by state

...we should come up with a single, modular system of styles and classes. Goals:

- [x] Consistency across all charts, both within charts themselves and parent containers for layout
- [X] Reduction in CSS bloat
- [x] `viewBox` only (no width and height attributes) on all `<svg>` elements to allow scaling and responsiveness across all breakpoints
- [x] Refined layout on smaller screens
- [X] Replacement of local/inline styling in JS component files in favor of reusable classes, such as y-axis stroke colors in [_AreaChart.js](https://github.com/COVID19Tracking/website/blob/d4d6def2a9754bdff0ef1a4e44a3998834dd94de/src/pages/dashboard/charts/_AreaChart.js#L60) and [_BarChart.js](https://github.com/COVID19Tracking/website/blob/d4d6def2a9754bdff0ef1a4e44a3998834dd94de/src/pages/dashboard/charts/_BarChart.js#L73)
- [ ] All chart classes moved to module SCSS file

### LINKED ISSUE #573: alignment of chart headers
let's align our chart headers to charts. here's how it should look like: 

Screen Shot 2020-04-15 at 12 05 01 PM

Screen Shot 2020-04-15 at 12 05 50 PM

Screen Shot 2020-04-15 at 12 07 26 PM

and let's also remove the text-center and add  margin-top: 20px; for the map * note, so it looks like this:
Screen Shot 2020-04-15 at 12 09 16 PM

### LINKED ISSUE #602: Dashboard has horizontal overflow on small viewports
On smaller viewports, we have horizontal overflow and scrolling as follows:

- Below 800px: due to the two column layout and non-responsiveness of "Total deaths in the US" and "Total Deaths By States" charts
- Below 400px: due to overflow with the map legend and tooltips

Recommendations:
- [X] Make "Total deaths in the US" and "Total Deaths By States" charts scalable/responsive (using `viewBox` only), stack vertically at smallest breakpoint. Best approach would create shared container styles to be used with the similar "Daily new tests in the US" two column layout
- [ ] Move the legend for the map so that it's vertically stacked with other elements on smallest viewports
- [ ] Disable tooltips on small screens, or move them to a static container as suggested by @nselikoff in https://github.com/COVID19Tracking/website/issues/454#issue-596946854

vijaysharwar:
In order to streamline things while making chart heading styles consistent, I've taken the step of refactoring the layouts overall. We now have three, reusable chart layouts with semantic class names:

- single chart (max width 500px)
- two column (horizontal layout for all breakpoints, max-width 250px per chart)
- two column large (stacks vertically on small breakpoints, horizontally for md and above, max-width 400px per chart)

In addition, this PR:
- Introduces dashboard module SCSS (but only for new layout classes)
- Removes extraneous layout classes/stylings
- Ensures all chart components use `viewBox` only to allow responsive scaling
- Adjusts the heading styles and alignment to fix #573
- Partially resolves #571 
- Partially resolves #602

---

#### Small viewport:
Screen Shot 2020-04-16 at 5 07 19 PM

#### Large viewport:
Screen Shot 2020-04-16 at 5 07 52 PM

---

#### Small viewport:
Screen Shot 2020-04-16 at 5 07 10 PM

#### Large viewport:
Screen Shot 2020-04-16 at 5 07 45 PM

---

#### Small viewport:
Screen Shot 2020-04-16 at 5 07 34 PM

#### Large viewport:
Screen Shot 2020-04-16 at 5 08 02 PM

---

UPDATE: this PR also adjusts the note below and adds a heading above the map (as per conversation on Slack with @julialedur).

Screen Shot 2020-04-16 at 5 26 21 PM


goleary:
oddly i'm seeing the long name on iphon x:
image
(after reload)

goleary:
something weird going on here:
image

goleary:
not sure why this is happening, but perhaps it's best to go ahead and add that resize listener?
also instead of using hardcoded 500, we should use the global constant from breakpoints.module.scss

vijaysharwar:
@goleary the `500` is because that was the point the title no longer fit on one line. I can use the next largest breakpoint `$viewport-ms` at `640px`, but it means showing `CTP` more often than needed. Still prefer that?

goleary:
500s fine.

vijaysharwar:
RE it not working when simulating iPhone, I'm seeing the same thing. I think it has to do with screen density. See https://stackoverflow.com/a/6908814. It works for me when I just make the viewport smaller. I'm going to try using a different native value.

vijaysharwar:
@goleary try now after 50273a1. Also had to resolve some conflicts. Please confirm I didn't nuke anything upstream with 53a0b2b.

goleary:
looks good and I manually checked the build.

Issue #620, opened 2020-04-16
N. Mariana Islands in sm multiples 


julialedur:
let's change Northern Mariana Islands to "N. Mariana Islands" in the sm multiples, so it doesn't break into two lines

also, let's use state abbreviations (NY, CA, etc) on sm screens to solve that problem. 

<img width="475" alt="Screen Shot 2020-04-16 at 6 01 15 PM" src="https://user-images.githubusercontent.com/39416238/79511852-1648f900-800e-11ea-840d-f99f36757a0d.png">

vijaysharwar:
Working on this now.

PR #621, opened 2020-04-16
fix(charts): prevent state names from breaking to new lines in small multiples

vijaysharwar:
- Changes "Northen Mariana Islands" to "N. Mariana Islands"
- Prevents text wrapping on small multiple state name links
- Use states keys on smaller screens

Fixes COVID19Tracking/website#620

Screen Shot 2020-04-16 at 6 38 30 PM
Screen Shot 2020-04-16 at 6 38 40 PM

### DIFF
--- a/src/pages/dashboard/_StateCumulativeTestsContainer.js
+++ b/src/pages/dashboard/_StateCumulativeTestsContainer.js
@@ -253,7 +253,14 @@ export default function CumulativeTestsByStateContainer() {
                   .toLowerCase()
                   .replace(/\s/g, '-')}`}
               >
-                <h4>{stateName}</h4>
+                <h4>
+                  <span className="small-multiples-chart-state-name">
+                    {stateName}
+                  </span>
+                  <span className="small-multiples-chart-state-key">
+                    {state.key}
+                  </span>
+                </h4>
               </a>
               <AreaChart
                 annotations={annotations}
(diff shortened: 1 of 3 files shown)


goleary:
is using the state key on small screens necessary?
I don't see any issue with the full state names on smaller screens:
image

jeremiak:
I agree @goleary - its in the original issue to use the postal abbreviations but if space allows for the human readable/full name of each state/territory we should do that. Postal abbreviations are common knowledge, but almost certainly not as common as the state names.

vijaysharwar:
@jeremiak @goleary I recommended switching to state code, as multiple states break onto another line at smaller breakpoints. Happens for me around 500px.

Screen Shot 2020-04-16 at 11 42 39 PM

vijaysharwar:
@jeremiak @goleary that said, another solution might be decreasing the number of items per row at the `$viewport-sm` breakpoint, from 4 to 3. Then going up to 4 per row at `$viewport-ms`. I can test that out more tomorrow if preferred to the state key solution.

vijaysharwar:
@jeremiak @goleary @julialedur see new PR https://github.com/COVID19Tracking/website/pull/633 that uses column change approach I mentioned above https://github.com/COVID19Tracking/website/pull/621#issuecomment-615024347.

vijaysharwar:
Closing in favor of solution found in https://github.com/COVID19Tracking/website/pull/633.

Issue #622, opened 2020-04-16
use same x-axis for all sm multiples


julialedur:
we need to use the same date range on the x-axis on the small multiple charts. 

@vijaysharwar: "We’ll need to calculate a shared extent across all states in _StateCumulativeTestsContainer.js and pass it to all AreaCharts ."

vijaysharwar:
I can work on this.

AaronMullan:
Well I have a simple solution for it, but usually Vijay has a more elegant one!  PR $#623

vijaysharwar:
I don't know if it's "elegant", but https://github.com/COVID19Tracking/website/pull/625 fixes this by passing a shared `extent` to all small multiple `AreaCharts`. We derive the value by checking all dates for all states. The benefit here is if some states dates don't match towards the most recent end of the scale, the charts will still have equivalent x-axes.

PR #625, opened 2020-04-17
fix(charts): use shared date extent for all small multiple charts

vijaysharwar:
The small multiples were each using an extent derived from the individual state's data. This means that their x-axes were not equivalent. It also caused the stay-at-home line in Puerto Rico to render outside its chart.

This PR uses D3 `extent` to find min/max dates across _all_ states, passing it to each `AreaChart` instance.

Fixes COVID19Tracking/website#622.

### Before

Screen Shot 2020-04-16 at 7 59 22 PM

### After

Screen Shot 2020-04-16 at 7 59 08 PM


src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -127,6 +128,16 @@ function groupAndSortStateDaily(query) {
   }
 }
 
+// TODO: we're iterating over data and calling parseDate multiple times in this
+// component. Seems like it could be optimized.
```
jeremiak, line 132: I agree, this could be optimized. We could do some of this work before we send it off to `groupAndSortStateDaily()` since the data that comes back from the query is actually all the data and we can just run `extent` with that. But I think its fine for now, especially since its only going to happen at build time.

vijaysharwar: Yup, so I started trying to do that, but ran into all kinds of headaches. There's `propTypes` to contend with in `AreaChart`, and the sorting function, and just a bunch of stuff. I figured this will do for V1 launch, and we can refactor later, prioritizing against other improvements.

PR #627, opened 2020-04-17
fix(dashboard): improve experience for screen readers in cumulative tests by state section (small multiples)

### LINKED ISSUE #610: sm multiples accessibility
as Todd Parker pointed out in this great thread, SVG charts aren’t useful to someone on a screen reader because they are just shapes, not readable data values, so he suggested:

- making the small SVGs “invisible” to a screen reader by adding focusable="false" aria-hidden="true" just to keep them from adding clutter as a person navigates through the grid (also hide toggle button and legend)

- adding a string of text that summarizes what the chart is trying to communicate visually by accessibly hiding it so it’s read by a screen reader but invisible in the layout. 
let's add: "This chart shows cumulative total and positive tests in all US states and territories. New York leads the testing race in both total tests and total tests per capita. When we look at absolute number of tests, New York is followed by California, Florida and Texas. But when normalized for population, the states rank changes significantly. In that scenario, New York is followed by Louisiana, Rhode Island and Massachusets."

nselikoff:
Specifically, this PR hides the small multiples and corresponding legend to screen readers until we can improve the accessibility of the AreaChart component further. Also adds "Sort by: " text to add context to the toggle between total and per-capita, which will clarify to all users what the toggle does.

Note that by default, if focusable and ariaHidden are not supplied to the AreaChart component, they will not be rendered in the HTML (i.e. both are "opt in").

Also note there is a visual change with the addition of "Sort by":

Screen Shot 2020-04-16 at 10 17 01 PM

Fixes #610


goleary:
I like what you've done here, but I wonder about the 'sort by' language.

That toggle not only changes the sort property, but also the display properties and "Sort by" doesn't feel like it fully capture that.

jeremiak:
I agree. What about some of these options (though I think in all cases we should either just sentence case the option or entirely lower case as if its part of the sentence:

* "Look at"
* "Display by"
* "Look at charts based on"

nselikoff:
@goleary @jeremiak great call. I like "Display by" personally, just added that and changed to sentence case in the latest commit:

image

src/pages/dashboard/_StateCumulativeTestsContainer.js
```diff
@@ -191,20 +191,25 @@ export default function CumulativeTestsByStateContainer() {
...
-          role="switch"
-          aria-checked={useTestsPerCapita}
-          tabIndex={0}
-        >
-          <span className={useTestsPerCapita ? '' : 'active'}>Total Tests</span>
-          <span className={useTestsPerCapita ? 'active' : ''}>
-            Tests Per Capita
-          </span>
+        <div className="dashboard-toggle-wrapper">
+          <div className="dashboard-toggle-label">Display by:</div>
```
vijaysharwar, line 195: This layout breaks on small screens. This is at iPhone X dimensions:

Screen Shot 2020-04-18 at 11 27 52 AM

vijaysharwar:
@nselikoff curious about rationale behind leaving some other surrounding elements.

https://github.com/COVID19Tracking/website/blob/e1398d1fc1f0ebd45e892643f07d1f5fffb94461/src/pages/dashboard/_StateCumulativeTestsContainer.js#L183-L211

Especially the toggle. Seems like if the charts are aria-hidden, then it doesn't make sense to have a screen reader detect those controls (there's nothing actionable for the user).

nselikoff:
@vijaysharwar good question - since the toggle changes the order of the listing of the states (and the state links will still be present to screen readers), my thought was that leaving it could still preserve some of the intended meaning (i.e. which states are "leading the race" in terms of testing when seen through those two lenses).

src/pages/dashboard/_StateCumulativeTestsContainer.js
nselikoff: @vijaysharwar thank you, just pushed an update for this (below screenshot at iPhone X dimensions):

image

vijaysharwar:
@nselikoff ah, that makes sense. So the titles are still displaying. In that case, a nice improvement might be to put the state multiples into an `<ol>` instead of `<div>` (suppressing the numbers with CSS) so that screen readers denote it as an ordered/ranked list.

And I think the SVGs could probably be made fairly screen reader accessible by proving ARIA labels. Like, maybe we could have a screen reader only text intro for each state's chart with the most recent totals/cumulative data.

Further reading/ideas: https://www.a11ywithlindsey.com/blog/accessibility-d3-bar-charts

nselikoff:
@vijaysharwar I like that idea re: `<ol>`. 

Also agree that an ARIA label for the small multiples would be better, and not difficult to highlight the most recent 2 data points + the stay at home annotation if it exists.

I'm wondering if the toggle should be marked as `role="radiogroup"` instead of switch (https://www.w3.org/TR/wai-aria-practices-1.1/#radiobutton) since we're really toggling between two interpretations, not switching a state on and off. Thoughts on that?

When I was reading about data viz accessibility the other day, in addition to the great link you shared, I also came across a good example from HighCharts of a line chart with a11y-focused construction, annotation and markup: https://www.highcharts.com/samples/highcharts/accessibility/accessible-line/. After v1 launches I think we should have a larger discussion on Slack about what we could do to further improve the AreaChart and other components.

vijaysharwar:
@nselikoff yeah, I've been curious about rationale behind `role="switch"` myself. It's not something I've used often, would need to read up on how various screen readers/browsers handle it. I don't know off top of my head if `"radiogroup"` would perform better.

However, there are as always lots of ways to skin this cat. I've been wondering if a [tab navigation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role) with `role="tab"` and `aria-controls` attributes might be best. That way, you can link the toggle button to the new content that is rendered.

None of this should hold up V1 launch, IMO. Probably worth merging this, and starting a new issue where we can define improvements over GItHub and Slack.

PR #630, opened 2020-04-17
Remove unused components, reorganize components

schwartzadev:


schwartzadev:
@kevee do you have any other components you'd like to add to the `utils` folder?

README.md
```diff
@@ -40,6 +40,16 @@ The site is now running at `http://localhost:8000`. Any changes you make to code
...
 Note that any changes you make while running Gatsby will automatically checked with [ESLint](https://eslint.org/), so check your console as you save files.
 
+### Organization
+
+Components live in `src/components` and are organized as follows:
+
+* Visualizations belong in `charts`
+* Components that are used more than once and those that are used across different parts of the website belong in `common`
+* Components that control the layout of the website (i.e. headers and footers) belong in `layout`
+* Components that are only used once or have a defined scope to a particular part of the website (i.e. `StateGrade` has to do with states) belong in their respectie directory in `pages`
```
goleary, line 50: spelling - respective

kevee: I would break this into a list with the directory first:

```markdown
* `/charts` - Data visualizations
* `/common` - Components that....
```

kevee:
I can't think of any more `util` components. Just a tiny change to the readme and I think we're ready to merge

README.md
schwartzadev: on it -- thank you both!

src/components/common/disclosure.js
```diff
@@ -1,21 +0,0 @@
-import React from 'react'
-import {
-  Disclosure,
-  DisclosureButton,
-  DisclosurePanel,
-} from '@reach/disclosure'
```
kevee, line 6: Since we're removing the only use of this package, let's remove it from `package.json` as well.

kevee, line 6: Let's remove this package as well

schwartzadev: fixed in e2dbf13a0aa2b7b70c233d2fa8f6d0cf8331676e

PR #631, opened 2020-04-17
fix: Blur jump-to-state input after hash change

### LINKED ISSUE #551: state picker doesn't work on mobile (chrome android)
the state picker doesn't work on mobile.

after having selected 'California':

image

error from console:
```
The FetchEvent for "https://covidtracking.com/data#state-ca" resulted in a network error response: an "opaqueredirect" type response was used for a request whose redirect mode is not "manual".
```

Android lts + Chrome lts

jasonlcrane:
This fixes #551, but there are other options outlined in that issue as well.


kevee:
Thanks, @jasonlcrane I'm building a preview and running tests in Browserstack to confirm. the `setTimeout` makes part of of me die as well, but I'm OK with that.

kevee:
So I tested this in Browserstack on a real device with Chrome and the error still showed up. I'll dig into this later today.

jasonlcrane:
Strange ... that preview works as expected for me on my Android phone (Chrome, Firefox and Edge), as well as on a a Galaxy phone on Browserstack. I'll see if I can break it.

PR #642, opened 2020-04-18
feat(charts): added a tooltip to the area chart

goleary:
leaving my work here for the night (or perhaps week).

- [x] needs updated styling
- [x] mobile behaviors needs to be lit up (only responds to mouseover currently)
- [ ] make sure it doesn't break storybook & that the default formatting makes sense outside of it's current use.

current style:
image

requested style:
image


goleary:
it also doesn't yet work for per capita. The small multiple chart container will need to pass a custom formatter for this.

goleary:
styling updated:
image

Updated to work for per capita.
It now functions on mobile tap, but the positioning of the tooltip is suboptimal on mobile. 
Some of this may be due to other issues on mobile.

julialedur:
that looks great, @goleary! 🎉 
my only feedback is i think the whole thing (box and font sizes can be like 20% smaller)
also, it doesn't seem to be working on safari? but maybe that's just a problem with the test link 

I will open another issue for responsiveness on mobile :)

src/pages/dashboard/_TotalAndPositiveLegend.js
```diff
@@ -0,0 +1,25 @@
...
+
+export default ({ date = null, positive, total, perCapita = false }) => (
+  <>
+    {date && (
+      <span className={clsx(styles.item, styles.date)}>{formatDate(date)}</span>
+    )}
+    <div className={styles.item}>
+      <span>{formatNumber(total)}</span>{' '}
+      <span className={clsx(styles.legendText, styles.total)}>
+        tests {perCapita && 'per million'}
```
vijaysharwar, line 15: I'm not sure if this is clear enough. Per million _what_? I'm tempted for it to be "tests per million people", but recognize that will be a lot of text to fit in the tooltip. Maybe smaller font size would make it possible...

Alternatively, maybe we footnote the "Tests per capita" text in the toggle, like we're doing with the "Stay at home" legend. So "per capita" is explained overall as a later aside. Then the labels in the tooltip could be "tests per capita" and "positive tests per capita". cc @julialedur

julialedur: @vijaysharwar yeah, I was thinking the same thing! 
I think the best solution is to do "tests per capita" on the tooltip
I think the note should be at the top though, cause it's an important information to know before you look at the graphic. 

so my suggestion is to either change the toggle button to say "Tests Per One Million People" (like the screenshot below) or to write "Per capita = per one million people" right below the toggle. I think the second option solves the long button issue, but is a bit repetitive

Screen Shot 2020-04-18 at 5 25 07 PM

vijaysharwar: At risk of nitpicking... I think I prefer keeping the toggle as is. It works best visually if the two sides are as equal width as possible. Plus, changing the width of the toggle means ensuring the layout doesn't break across viewports, especially small/medium screens. Don't forget that @nselikoff added "Display by:" text in front of the toggle as well, via https://github.com/COVID19Tracking/website/pull/627#discussion_r410715402. So we're tight for horizontal space at the point that the legend goes onto same line as the toggle.

Elsewhere, in the map section, we're doing it this way:

> Per capita*
> 
> \* Per one million people

Might we be consistent? Use "Tests Per Capita*" text in both the toggle and tooltips. Then put a note below with the other footnote. It's findable if people _need it_, but otherwise doesn't interfere with usability and layout. Like so:

Screen Shot 2020-04-18 at 7 00 50 PM

vijaysharwar:
@goleary here's a rough proof of concept for preventing the tooltips from getting squished or going off the right edge of the screen: https://github.com/vijaysharwar/website/commit/b5f70c94c625c291354c341fd4304ef723a9253e

Downside is it requires setting a fixed width in the tooltip styles. But I couldn't figure out how to get the width of the rendered tooltip ahead of time; bit of a chicken/egg dilemma.

Same approach could be applied to `y` and `top` values to prevent going off the bottom edge of the screen.

I really like that you made a `Tooltip` component. Would be great to eventually have the map and any other chart use the same (allowing parent to pass in styles and children layout). Consolidate our tooltip positioning/behavioral logic across all uses.

src/pages/dashboard/_TotalAndPositiveLegend.js
julialedur: yes! that's great!

goleary:
i quite like your approach @vijaysharwar, I think i'll steal that :)

vijaysharwar:
@goleary checking latest, I think with the fixed width tooltips we can tweak the styles when the text gets too long. A little smaller font size, and maybe greater line-height.

Screen Shot 2020-04-20 at 2 47 08 PM

Also I think we should tweak the text to just say "tests per capita"/"pos. tests per capita", and implement footnotes like discussed here:

https://github.com/COVID19Tracking/website/pull/642#discussion_r410767278

cc @julialedur

julialedur:
agreed! let's make the fonts smaller and change it to "per capita"

PR #641, opened 2020-04-17
fix(dashboard): Fix y axis styling on bar & x axis on horizontal bar

jeremiak:
After:

Screen Shot 2020-04-17 at 6 31 09 PM
Screen Shot 2020-04-17 at 6 31 16 PM

Fixes #639

PR #645, opened 2020-04-18
fix(charts): fix regression issues with charts layout and responsiveness

### LINKED ISSUE #644: Dashboard responsive charts layout issues on iPhone
We're still having issues with responsive charts layout on iPhone X. See screenshots below.

Screen Shot 2020-04-18 at 10 42 15 AM

Screen Shot 2020-04-18 at 10 42 41 AM

vijaysharwar:
Looks like some merge conflict resolution snuck older markup back. In this PR:

- Removed width/height attributes in `BarChart` svg
- Removed unnecessary wrapper divs around `BarChart` and `HorizontalBarChart`

Fixes #644. Regression issues will be fully resolved with the additional merge of https://github.com/COVID19Tracking/website/pull/641.

## Before

Screen Shot 2020-04-18 at 10 42 15 AM

Screen Shot 2020-04-18 at 10 42 41 AM

## After

Screen Shot 2020-04-20 at 2 59 43 PM

Screen Shot 2020-04-20 at 3 00 17 PM

### DIFF
--- a/src/components/charts/bar-chart.js
+++ b/src/components/charts/bar-chart.js
@@ -43,68 +43,61 @@ const BarChart = ({
   )
 
   return (
-    <div align={align}>
-      <svg
-        className={chartStyles.chart}
-        width={width}
-        height={height}
-        viewBox={`0 0 ${width} ${height}`}
-      >
-        <g transform={`translate(${marginLeft} ${marginTop})`}>
-          <g>
-            {yScale.ticks(yTicks).map((tick, i) => (
-              <g key={tick}>
-                {i < showTicks && (
-                  <>
-                    <text
-                      className={chartStyles.yTickLabel}
-                      y={yScale(tick) + 6}
-                      x={`${tick}`.length * -12}
-                    >
-                      {formatNumber(tick)}
-                    </text>
-                    <line
-                      className={chartStyles.gridLine}
-                      x1={0}
-                      x2={width - totalXMargin}
-                      y1={yScale(tick)}
-                      y2={yScale(tick)}
-                    />
-                  </>
-                )}
-              </g>
-            ))}
-          </g>
-        </g>
-
-        <g transform={`translate(${marginLeft}, ${height - marginBottom})`}>
-          {ticks.map(d => {
-            const date = xScale.domain()[d]
-            return (
-              <text
-                className={`${chartStyles.label} ${chartStyles.xTickLabel}`}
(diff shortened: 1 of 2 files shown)


src/components/charts/bar-chart.js
```diff
@@ -43,68 +43,61 @@ const BarChart = ({
   )
 
   return (
-    <div align={align}>
```
jeremiak, line 46: I had not idea that `align` was even a valid attribute for `<div>` elements! Did a bit of research and it looks like [MDN/caniuse](https://caniuse.com/#feat=mdn-html_elements_div_align) says that this attribute is deprecated and shouldn't be used. Noting for posterity

src/components/charts/horizontal-bar-chart.js
```diff
@@ -31,57 +31,55 @@ export default function HorizontalBarChart({
...
-        </g>
-      </svg>
-    </div>
+          </g>
+        ))}
+      </g>
+
+      <g transform={`translate(0, ${marginTop})`}>
+        {data.map(d => (
+          <svg
```
jeremiak, line 58: Why embed another `<svg>` inside of the existing `<svg>`?

src/components/charts/bar-chart.js
vijaysharwar: Yeah, I didn't think we should use it, and the layout is now handled by classes with flexbox.

src/components/charts/horizontal-bar-chart.js
vijaysharwar: Having the nested element allowed me to keep the containers left aligned, but text right justified with `textAnchor="end"`. Was [following example here](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-anchor).

From what I've read, nesting `<svg>` is allowable, and helps mimic relative positioning:

https://www.sarasoueidan.com/blog/nesting-svgs/
http://tutorials.jenkov.com/svg/svg-element.html

I tried using a `<g>` element but it clobbered the layout, stopped honoring the y position, and I couldn't figure out why:

Screen Shot 2020-04-20 at 4 37 05 PM

Since `<svg>` worked, I left it at that. But welcome other ideas!

jeremiak: Oh, woah! Thanks for pointing me towards this, that's a pretty powerful technique I had no idea about.

Issue #496, opened 2020-04-10
Feature request: Volunteer Credits page form 


amandafrench:
Please let us know the page you are talking about, and your content suggestions!

About the Project at https://covidtracking.com/about-project has a list of contributors at the bottom -- currently that is updated manually from a Google Sheet. We'd like it to be an easier process than that: @kissane suggested that the existing code for press links could be reused for this purpose. Volunteers do need to be able to opt in to the credits, so no mass harvesting without permission, but other than that I have no specs.

amandafrench:
However I do have OPINIONS, one of which is maybe that should be a whole Separate Page now that our community is groooooowing. kthxbye

kevee:
I'm thinking we can use the Slack API to just read all our volunteers from Slack. How about "if you join the channel #put-me-on-the-website you will be automatically added?

kissane:
You know, I…do not hate that.

kissane:
Whatever is easier for volunteers to do, which it seems like Slack channel is probably easier.

amandafrench:
Man, I feel like I should hate the proliferation of Slack channels more than I do, but they are damn useful.

amandafrench:
Would that method allow people to put not just their names but also their links?

kevee:
We could just use their profile info from Slack.

kevee:
I did a very short search and here's some thoughts:

There's already a few plugins to pull slack users into Gatsby. The technical part should be easy.

We can add new profile fields for public website, public name, and public team membership. 

They can add themselves to the public website through either:
Users join a special channel
Or
They just select a "make me public" custom field in their profile.

The latter is easy to explain, the former lets us administer membership more easily, since folks can be removed from the channel or added if needed by admins

muamichali:
I am going to work on this! Starting with a test slack instance until I have it down. Will invite you @amandafrench  and @kissane for testing soon.

kevee:
Quick note: This will require creating a new API integration with `users:read` scope.

kevee:
**Crazy idea # 4722**: In the past, I've always tried list numbers of volunteers to illustrate engagement & impact. What if we had this page lede with "We have [x number] volunteers, [y number] of whom have been active in the past week. Here are some of those:"

amandafrench:
I'd love that! Have been reporting Slack members & guests in #status from the Slack member directory. Be great if you have some way of measuring "active" volunteers too (e.g., who has posted anything in Slack in the past week).

[...]

PR #647, opened 2020-04-18
feat(about): add volunteers page

muamichali:
Continuing work on https://github.com/COVID19Tracking/website/issues/496


src/components/common/volunteers-list.module.scss
```diff
@@ -0,0 +1,3 @@
+.list {
```
muamichali, line 1: I wasn't sure if we need a unique name or not here.

kevee: Nope! The classnames in modules get scoped to their modules.

muamichali:
build error

The build process couldn't find _/data/v1/volunteers.json 

Did I miss a step to add it to the archive?

kevee:
The fix you made earlier today is now in Netlify, so we are live at:

https://deploy-preview-647--upbeat-lovelace-3e9fff.netlify.app/about-project/volunteers

muamichali:
- [ ] We have to stack the columns in mobile...

sarawb:
Got a couple logic issues for you! 

- There's duplication between leads and contributors list; we shouldn't show in contributors if they're already in leads. 
- Leads is alphabetized by last name, contributors by first name. No strong opinions about which we use but I'd like both to be ordered by the same thing. Guessing FN all around will be easier, since we don't use separate FN/LN fields in Slack for this. 

(Design review still tk)

sarawb:
Oh, just noticed that links for users who did not enter http:// before their websites in Slack aren't working. Can we automatically append that? Large and changing volunteer base is gonna make this hard otherwise. Examples: Aviva, Bree, Gabe, Marcie (not a complete list).

kevee:
@sarawb - The main contributors list is automatically pulled from Slack, while the leads list is content from Contentful. I think it's cleaner for us to just ask the leads to disable their listing in Slack, then we can manually manage their public presence in Contentful.

schwartzadev:
>     * [ ]  We have to stack the columns in mobile...
Fixed in 763ec76c7b531ae525c0d1c69613085a4dc47545

> Oh, just noticed that links for users who did not enter http:// before their websites in Slack aren't working. Can we automatically append that? Large and changing volunteer base is gonna make this hard otherwise. Examples: Aviva, Bree, Gabe, Marcie (not a complete list).

Fixed in e479eeb6322593e176a719a6803f57753b0a8396

schwartzadev:
> Leads is alphabetized by last name, contributors by first name. No strong opinions about which we use but I'd like both to be ordered by the same thing. Guessing FN all around will be easier, since we don't use separate FN/LN fields in Slack for this.

Great call! Yes, sorting by first name would definitely be easier. I'll look into sorting the leads that way in contentful.

kevee:
We're holding off on publishing this PR for a few days to give people time to sign up in slack.

beep:
Design-wise, would it be possible to disable the bullets on the contributor list, once we move to multiple columns? I think they’re fine in the single column display, but they look a bit busy on the wider layouts. I might also suggest bringing the line-height down a little bit, maybe to 1.3.

careeningspace:
I noticed that I am on the page twice and Prajakta is not listed as a Lead. Does Prajackta need to edit something in her profile?

schwartzadev:
@careeningspace see Kevin's comment:
> I think it's cleaner for us to just ask the leads to disable their listing in Slack, then we can manually manage their public presence in Contentful.

That way the leads won't appear in the main list below. I think that we'll plan to do this before we merge.

Blocking merge :warning:
-----
- [ ] Team leads need to remove their "Yes I want to be on the website" field on their Slack profiles so they no longer appear in the main contributor section

schwartzadev:
> 
> 
> Design-wise, would it be possible to disable the bullets on the contributor list, once we move to multiple columns? I think they’re fine in the single column display, but they look a bit busy on the wider layouts. I might also suggest bringing the line-height down a little bit, maybe to 1.3.

@beep fixed in 7dc5be2 and 6895ebd :tada:

beep:
HECK YEAH @schwartzadev, thank you for doing that!

PR #660, opened 2020-04-20
feat(sheets): handle Google Sheets HTTP errors

schwartzadev:
This adds a more clear error for whenever fetching Google Sheets fails

### DIFF
--- a/build/handlers/index.js
+++ b/build/handlers/index.js
@@ -1,3 +1,4 @@
+const util = require('util')
 const _ = require('lodash/fp')
 const { getSheet } = require('./sheets')
 const getXml = require('./xml')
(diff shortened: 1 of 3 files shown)


webmasterkai:
Instead of having the error handler inline I broke it out into its own function called `rejectOnError`. Instead of throwing an error within sheets.js I made the error handling happen inside `fetchParseFix()` within index.js. The final process exit happens in `exitError()` that includes code to print in red and a process.exit(1) to avoid `UnhandledPromiseRejectionWarning` messages.

Before the message was
```javascript
{
  error: {
    code: 400,
    message: 'API key not valid. Please pass a valid API key.',
    status: 'INVALID_ARGUMENT',
    details: [ [Object] ]
  }
}
``` 

Now the error prints as 
```
API ERROR
Google Sheets error. HTTP code: 400 (API key not valid. Please pass a valid API key.)
```

PR #664, opened 2020-04-21
Add GraphQL support to website using Netlify functions

ccheever:
This PR will add working GraphQL support, including an interactive GraphQL playground to the website using Netlify functions.

### Source and development

The source for the GraphQL stuff lives under `src/graphql`.

You can read the README.md in that directory for instructions on how to develop and test it.

### Data

The data for this API is pulled from the production v1 JSON API. This could be changed to be pulled from static JSON files in the future, once that is setup.

The thing that is nice about this is that as long as the v1 JSON API is consistent, the GraphQL API shouldn't break in surprising ways.

### Netlify Stuff

A new script (`build:graphql`) is added to the main `package.json`.

At build time, this creates a directory `src/graphql/functions-dist` where the Netlify Cloud functions live.

The `build:graphql` script is added to the main `build` script that runs all the build subscripts.

A `functions` directive is added to the `netlify.toml` configuration specifying where the production functions code lives.

```toml
  functions = "./src/graphql/functions-dist"
```

### Testing this PR

In general, you can visit the root production URL plus `.netlify/functions/graphql` to visit the GraphQL playground. This doubles as the graphql endpoint.

A test deploy of this repo with this PR added is currently here:
https://pedantic-mclean-da3736.netlify.app/.netlify/functions/graphql

An example query you could try is:

```graphql
query {
  usCumulativeTotal {
    deaths
  }
}
```

or a more complicated one could be:

```graphql
query {
  state(state:"PA") {
    name
    cumulativeData {
      total
      positives
      negatives
    }
    notes
    puiReporting
    pumReporting
  }
}
```

### Future Work

It may make sense to pull the data from another source besides the production JSON API. For example, it might make sense to require statically built JSON files if they are included in the repo.

It may make sense to add additional fields and things like that. This was up-to-date about 2 weeks ago, but things like hospitalization have been added since then.

### GraphQL Schema

To see the schema used here, take a look at the `typeDefs.js` file in this diff or use the interactive schema browser from the GraphQL playground.


kevee:
Pausing on this while we get clarity from Netlify on our plan and quotas

webmasterkai:
If I have a moment I'll go through and update the json endpoints being used. It'd be great to update them to the latest https://covidtracking.com/api/v1/* endpoints instead of the old ones.

- /api/cdc/daily -> /api/v1/cdc/daily.json
- /api/press -> /api/v1/press.json
- /api/states -> /api/v1/states/current.json
- /api/states/daily -> /api/v1/states/daily.json
- /api/states/info -> /api/v1/states/info.json
- /api/states/grade -> /api/v1/states/grades.json
- /api/us -> /api/v1/us/current.json
- /api/daily -> /api/v1/us/daily.json
- /api/counties -> /api/v1/counties.json
- /api/urls -> /api/v1/urls.json
- /api/screenshots -> /api/v1/states/screenshots.json

src/graphql/dataAsync.js
```diff
@@ -0,0 +1,164 @@
...
+  // The `rawDay` value is typically an int not a string, of a form like 20200317
+  try {
+    return luxon.DateTime.fromFormat(rawDay + '', 'yyyyMMdd').toJSDate();
+  } catch (e) {
+    console.error(`Failed to parse raw day '${rawDay}'`);
+    return null;
+  }
+}
+
+function rawWeirdEasternTimeToJSDate(rawDateTimeStr) {
```
kevee, line 35: It really is weird.

kevee:
Agree with @webmasterkai - this should be hitting the new API endpoints. But you could even do one better:

Since the site is already building JSON files into `./_data` on build, and you're running a build command for functions, you could copy all those data artifacts over and let the function just read the data off the local filesystem.

ccheever:
Closing this and replacing with #689

PR #671, opened 2020-04-21
Build handler unit tests

gitjeff05:
This adds a bunch of tests for the `build/handlers/index.js` module.

### DIFF
--- a/build/handlers/index.js
+++ b/build/handlers/index.js
@@ -26,12 +26,12 @@ function defaultPage(path) {
   ]
 }
 
-function fetchParseFix(resource) {
+function fetchParseFix(resource, processResultFn = processResult) {
   const { app, fetch } = resource
   const handler = fetch || handlers[app]
   if (!_.isFunction(handler))
     throw new Error(`Handler not found: ${resource.app}`)
-  return handler(resource).then(processResult(resource.fixItems))
+  return handler(resource).then(processResultFn(resource.fixItems))
 }
 
 function exitError(err) {
@@ -41,15 +41,18 @@ function exitError(err) {
   process.exit(1)
 }
 
-function fetchSave(resource) {
+function fetchSave(resource, fetchParseFixFn = fetchParseFix) {
   const save = resource.createPages || defaultPage(resource.path)
-  return fetchParseFix(resource)
+  return fetchParseFixFn(resource)
     .then(save)
     .then(saveAll)
     .catch(exitError)
 }
 
 module.exports = {
+  processResult,
+  defaultPage,
   fetchParseFix,
   fetchSave,
+  handlers,
 }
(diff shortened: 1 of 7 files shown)


build/handlers/index.js
```diff
@@ -26,12 +26,12 @@ function defaultPage(path) {
   ]
 }
 
-function fetchParseFix(resource) {
+function fetchParseFix(resource, processResultFn = processResult) {
```
gitjeff05, line 29: Passing this in as a default. This way I can mock it in the tests. Since I can't really run `build:fetch` without the keys, I was hoping @webmasterkai or @kevee could have a second look at this. I don't think it will cause any problems.

build/handlers/index.js
```diff
@@ -41,15 +41,18 @@ function exitError(err) {
   process.exit(1)
 }
 
-function fetchSave(resource) {
+function fetchSave(resource, fetchParseFixFn = fetchParseFix) {
```
gitjeff05, line 44: Same comment as in `fetchParseFix`

build/handlers/index.js
kevee: We actually don't do the API build in CircleCI but instead [download the data](https://github.com/COVID19Tracking/website/blob/master/.circleci/config.yml#L21) using `npm run setup`.

gitjeff05: Ah, okay. I'm a little confused about our CI and build processes then. Where and when do the steps in `npm run build` get run? I kind of assumed CircleCI did testing and deployment.

gitjeff05: Oh, I think I get it now. I did not look closely enough at the CircleCI config. I assumed it was doing the entire `npm run build`. So, the `build:fetch` and `build:data-archive` steps are still being done elsewhere.

build/handlers/index.js
gitjeff05: I could of course make manual mocks for these the same way I did for `./build/handlers/save` but I believe then we'd have to move these out into their own module. I don't think it is a huge deal though.

gitjeff05:
@kevee That all sounds reasonable. So, here is what I'm thinking for the file structure (I should probably make a note in the README.md):

- `src/__tests__/build/` **tests that website built correctly**
- `build/__tests__/` **tests for code run by `build:fetch`**

And in the jest config file, we will explicitly ignore:

- `./src/__tests__/build/index.js`
- `./build/test.js`

This way, `npm run test` will run everything -- but **not** the website build tests. And `npm run build:test` will just run the website build test but not these newly added tests.

kevee:
Circleci was unhappy during a Github outage. I [reran and the tests passed](https://app.circleci.com/pipelines/github/COVID19Tracking/website/234/workflows/c3e798db-2e7e-40d9-9d39-fbfe00bf7b23).

kevee:
Testing worked, just didn't report - https://app.circleci.com/pipelines/github/COVID19Tracking/website/251/workflows/51533b75-3877-46ec-94c1-468618df731f

PR #638, opened 2020-04-17
Layout fixes; consolidating typesetting into SCSS; refreshed Visualization Guide

beep:
Hey folks! Got a big ol’ PR here, which I’d love to treat as a _discussion starter_ more than anything. There’s a lot that’s changed here, and I’d love to get as many thoughts as possible before (and if) we fold it in.

Basically, I had two goals for opening this up:

1. To do a visual QA/refinement pass on our v2 redesign;
1. To consolidate type styling and layout control in our SCSS. (And—hopefully!—retire [reflexbox](https://rebassjs.org/reflexbox/) and [typography.js](https://kyleamathews.github.io/typography.js/) as a result.)

This touched a fair bit of the codebase, so I’d appreciate any thoughts folks have on this. And I’d _really_ appreciate some design review as well, as I’ve tried tightening up a few things throughout the site. (Nothing major, mind—mainly some more consistent margins and heading treatments—but still! Feedback! I love it!)

Additionally, per some discussion below, we now have a new `featured` layout for our Visualization Guide. Thanks so much to @jasonsantamaria and @julialedur for the design feedback!

Code-wise, there’s one additional item I snuck in: I renamed `<GetInvolved />` to `<ListArrow />` in https://github.com/COVID19Tracking/website/pull/638/commits/3d14a6ac019a8f939865afccc765c5a80c9d8120. This is _totally_ out of scope for this PR, but _also_ please treat it as a proposal! It seemed like a more generic name for this component might be helpful since we’re starting to use it in more places. But as with everything else above, I’d be more than happy to discuss/revert this if others would prefer.

Thanks for your feedback, all!

PR #682, opened 2020-04-22
fix(typography): restore box-sizing globally

vijaysharwar:
## Description

Removing Typography.js seems to have eliminated global border-box sizing, which negatively affects several layouts in the Visualization Guide.

Setting box-sizing to border-box changes how the browser calculates width of elements, including their paddings in the final width calculation.

So if you have an element with width of 100px, and 20px left/right padding, the browser will render it at 100px with 60px of content area. Without border-box, that same container would be rendered to 140px total.

## Related Issues

See comment in https://github.com/COVID19Tracking/website/pull/638#issuecomment-617871028.

### Before:

Screen Shot 2020-04-22 at 11 50 37 AM

Screen Shot 2020-04-22 at 11 54 11 AM

### After:

Screen Shot 2020-04-22 at 11 50 46 AM

Screen Shot 2020-04-22 at 12 26 07 PM

### DIFF
--- a/src/scss/global.scss
+++ b/src/scss/global.scss
@@ -2,9 +2,21 @@
 @import './colors.module.scss';
 @import './breakpoints.module.scss';
 
+html {
+  box-sizing: border-box;
+}
+
+*,
+*:before,
+*:after {
+  box-sizing: inherit;
+}
+
 body {
   color: $text;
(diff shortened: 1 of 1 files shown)


schwartzadev:
LGTM -- merging to @beep 's branch :+1:

PR #689, opened 2020-04-22
[GraphQL] Setup GraphQL using Netlify functions and static json files

ccheever:
This PR will add working GraphQL support, including an interactive GraphQL playground to the website using Netlify functions.

## Description

The overarching theme of this implementation is to create something
that we can maintain with low effort. We use Netlify functions so that
this is part of the same repo and deployment cycle as the rest of the
website and API (and same billing account, etc., etc.). The GraphQL API
uses the JSON API v1 as its datasource, so as long as that is maintained,
this should still work.

If we want to surface additional data in the future, for example race data,
via this GraphQL service, that will require additional work. I can't think
of any good way around that right now, but I think it is OK. Adding additional
data here should be as easy as just making some basic modifications to the
`dataAsync.js` data fetching file, `typeDefs.js` schema, and `resolvers.js`
resolvers.

### Source and development

The source for the GraphQL stuff lives under src/graphql.

You can read the README.md in that directory for instructions on how to develop and test it.

### Data

The data for this API is pulled from the production v1 JSON API. This could be changed to be pulled from static JSON files in the future, once that is setup.

The thing that is nice about this is that as long as the v1 JSON API is consistent, the GraphQL API shouldn't break in surprising ways.

To see the schema used here, take a look at the typeDefs.js file in this diff or use the interactive schema browser from the GraphQL playground.

### Netlify Stuff

A new script (`build:graphql`) is added to the main `package.json`.

At build time, this creates a directory `src/graphql/functions-dist` where the Netlify Cloud functions live.

The `build:graphql` script is added to the main build script that runs all the build subscripts.

A functions directive is added to the `netlify.toml` configuration specifying where the production functions code lives.

```toml
  functions = "./src/graphql/functions-dist"
```

### Testing this PR

In general, you can visit the root production URL plus `.netlify/functions/graphql` to visit the GraphQL playground. This doubles as the graphql endpoint.

A test deploy of this repo with this PR added is currently here:
https://pedantic-mclean-da3736.netlify.app/.netlify/functions/graphql

An example query you could try is:

```graphql
query {
  usCumulativeTotal {
    deaths
  }
}
```

or a more complicated one could be:

```graphql
query {
  state(state:"PA") {
    name
    cumulativeData {
      total
      positives
      negatives
    }
    notes
    puiReporting
    pumReporting
  }
}
```

If we put this into production, we would likely want to add some automated tests and monitoring.
In the short term, doing a manual test using the playground seems the most sensible way to test with confidence.

### Future Work

It may make sense to add additional fields that we are now tracking things like that. This was up-to-date about 2 weeks ago, but things like hospitalization have been added since then. As we add things like race data, we may want to expose them as well. Some fields may be going away, such as posNeg (postitive + negative). We may want to remove those now before rolling out to production.

## Related Issues

This replaces PR #664 which got gnarly because of some conflicts.


webmasterkai:
Do we have a script setup to auto-fix the lint errors?

ccheever:
> Do we have a script setup to auto-fix the lint errors?

I can take a look at just manually fixing them for now. They just look like small things.

src/graphql/stateNames.js
```diff
@@ -0,0 +1,83 @@
+/* eslint-disable */
+
+let names = {
```
kevee, line 3: Can this come from the API data?

Nosferican:
Any updates on this?

ccheever:
@Nosferican should land in the next few days

Nosferican:
Perfect. Is it safe to assume that it will be first tier and support the latest RESTful schema? Trying to evaluate if I should wait a few days and use everything GraphQL or update one more iteration with the REST API.

ccheever:
@Nosferican In terms of service uptime, data quailty, etc., the GraphQL API can be considered first class once it launches.

One caveat: The GraphQL layer will be built on top of the data that is used for the JSON API. If new data is surfaced by the project, there may be a window of time where the data is available via JSON but not via GraphQL. I'll try to keep those windows of time small.

Nosferican:
Bump.

PR #690, opened 2020-04-23
fix(summary-table): DOMNesting

### LINKED ISSUE #680: Our data: validateDOMNesting(...): <col> cannot appear as a child of <table>
**Describe the bug**
When navigating to the Our data page (in development mode), a console error is printed repeatedly: 
`validateDOMNesting(...): <col> cannot appear as a child of <table> `

This happens in `components/common/summary-table.js:20-21`:  
```javascript
<Table 
//  ...
  <col />
  <col />
```

I guess using a colgroup for those would do, but I'm not comfortable enough with this part of the codebase to give it a try without breaking all the tables (and there are many).

This **doesn't break the build** but I guess we want valid HTML (and flooding the console can obfuscate more problematic issues).

schwartzadev:
## Description
Fixes invalid DOM issues with the `<SummaryTable>` component

Summary tables without outcomes:
image

Summary tables with outcomes:
image

This also updates `package-json`.

## Related Issues
closes #680 

## Go-live time
n/a


src/__tests__/components/pages/data/__snapshots__/state-list.js.snap
```diff
@@ -200,7 +200,7 @@ exports[`Components : Pages : Data : State data renders correctly 1`] = `
       <div
         className="detail-text detailText"
       >
-        Last updated: Fri Apr 17 2020 12:00 am ET
```
kevee, line 203: Looks like we have a timzeone issue in the test. I'll fix that.

PR #692, opened 2020-04-23
Issue 688: Added margin to the press logos

### LINKED ISSUE #688: margin after press logos
this section of the website feels a bit crowded. 
i suggest we add a breathing space between the press logos and the list of stories. 

Screen Shot 2020-04-22 at 4 45 22 PM

karinanguyen:
## Description

Added margin for press logos (Issue #688) 

## Related Issues
Fixes #688


src/components/pages/homepage/press-logos.module.scss
```diff
@@ -8,5 +8,6 @@
     filter: grayscale(1);
     height: 40px;
     margin: 1rem 2rem;
+    margin: 3rem;
```
kevee, line 11: Hi, @karinanguyen, and thanks for your first PR! I think we shouldn't have two `margin` declarations in here.

karinanguyen: So sorry, thank you for the catch ! Just edited:)

src/components/pages/homepage/press-logos.module.scss
```diff
@@ -8,5 +8,6 @@
     filter: grayscale(1);
     height: 40px;
     margin: 1rem 2rem;
+    margin-bottom: 3rem;
```
schwartzadev, line 11: Thanks for this, @karinanguyen ! I think what @julialedur was hoping for in #688 was to add margin between the article titles and the press logos, where the two grey lines are in this pic:
image

I _think_ you could do that by applying the same patch to `.logos` (the container) instead of `img` (the children).

PR #694, opened 2020-04-23
fix(dashboard): re add accidentally deleted name (mine!)

goleary:


goleary:
@kevee whoops, that's what I get for making a small text edit online.
Thanks @schwartzadev for fixing it up.

With another review we can merge and close this out.

Issue #693, opened 2020-04-23
feat: Search


kevee:
We need to add a search feature to the website. I [started a branch on this](https://github.com/COVID19Tracking/website/tree/kevee/algolia-search), but never finished. Please start a new branch called `feature/search` so we can do PRs against it.

Some big things to consider: Algolia breaks different types of documents into different "indices" - I made one for states, blog posts, and pages, and prefaced them with `test_`. Eventually we can create a separate `live_` version of these indices.

In the branch, I added a plugin for Algolia, and Gatsby updates the search index *on every build* so we get the most updated data in the indexes. So it basically updates all states, pages, and blog posts for every deploy. You can [see how I setup configuration here](https://github.com/COVID19Tracking/website/blob/kevee/algolia-search/src/utilities/algolia.js). It's a bit rough, and I feel like dumping entire page and blog post bodies into Algolia is the wrong way to go - maybe there's a way to just put in like keywords? Unsure.

The index name prefixes (i.e. `test_` should be determined by an environment variable so we can different ones for different environments.

The public search API key (read only) [should be available to the search interfaces](https://www.gatsbyjs.org/docs/environment-variables/#client-side-javascript) by using a `GATSBY_` prefixed environment variable.

What we need is a search interface that allows:

- [x] Ordering by index type, i.e. show states first, then blog posts, then pages - if I type "California" I probably want to jump to California's page
- [x] A search results page at `/search` that supports queries - i.e. `/search?q=stuff`
- [x] A search box that goes on the header. Do *not* worry about design right now, but just make sure it works.
- [x] A search automcomplete - we already use [Reach's combobox](https://reacttraining.com/reach-ui/combobox) component on the site for the state dropdown, and I like it because of all all the accessibility features in it, but I am not wedded to that component

EDIT: Adding the few last items discussed on Slack, please add if I'm missing something: 
- [x] Add last update date for States in index + results  
- [x] Add last update date for Pages in index + results  
- [x] Add States `notes` to index

superbiche:
@kevee should we keep the todo here or on the PR? For now I'm updating the PR one

muamichali:
**Idea**: We could index our tweets as well and display them on the site as search results. A lot of our content so far is in tweets. CC @kissane for your thoughts

superbiche:
@muamichali if you have an API key for the Twitter account I'd be happy to give it a try so we can have something to iterate on

PR #700, opened 2020-04-23
feat(search)

superbiche:
## Description
Draft PR to add Algolia search (#693) 

**Original todo from @kevee**, please add items here so we don't get lost :) :
- [x]  Ordering by index type, i.e. show states first, then blog posts, then pages - if I type "California" I probably want to jump to California's page
- [x]  A search results page at /search that supports queries - i.e. /search?q=stuff
- [ ]  A search box that goes on the header. Do not worry about design right now, but just make sure it works.
- [ ] A search autocomplete - we already use Reach's combobox component on the site for the state dropdown, and I like it because of all all the accessibility features in it, but I am not wedded to that component
- [x] Environment-aware Algolia index (https://covid-tracking.slack.com/archives/C012DPFRP7D/p1588079930004600)  
- [x] Target the `feature/header-refresh` branch  
## Related Issues
#693


muamichali:
I think it could be useful to scrape our own tweets and index them as well, since a lot of useful information/thoughts/explanations happen on our twitter feed.

superbiche:
@muamichali yup that's something we can work on, if someone's willing to create the app and send us the API keys on Slack, we'd be happy to give it a shot when we're done with the basic features!

src/pages/search.js
```diff
@@ -0,0 +1,146 @@
...
+        <div>
+          <img
+            src="https://d1j8pt39hxlh3d.cloudfront.net/products/previews/RES3POBSZ353HFVPZOKR/2329_kKbUTLGEVXhuOIJqrT7QvIJFAXvEpQ3z.gif"
+            alt="Searching..."
+          />
+          <h3>Searching...</h3>
+        </div>
+      )}
+      <div className={searchPageStyle.searchResults}>
+        {/* State results */}
```
schwartzadev, line 84: These could probably be done DRY-er -- we could have a component for the results and then pass the inner content to the component for:
* state results
* blog posts
* pages

Plus, this should make it easier to add items like Twitter in the future!

src/pages/search.js
```diff
@@ -0,0 +1,146 @@
...
+            await querySearch(searchState, searchDispatch)
+          }}
+        >
+          Search
+        </button>
+      </form>
+      {searchState.isFetching && (
+        <div>
+          <img
+            src="https://d1j8pt39hxlh3d.cloudfront.net/products/previews/RES3POBSZ353HFVPZOKR/2329_kKbUTLGEVXhuOIJqrT7QvIJFAXvEpQ3z.gif"
```
schwartzadev, line 77: This is a placeholder for now -- just flagging it before we push

src/utilities/algolia.js
```diff
@@ -0,0 +1,139 @@
...
+        slug
+        body {
+          body
+        }
+      }
+    }
+  }
+}`
+
+function splitBodyIntoChunks(baseChunk, firstChunk, bodyChunks) {
```
schwartzadev, line 55: I assume this is based on Agolia limits? Could you add a docstring to clarify?

src/pages/search.js
```diff
@@ -0,0 +1,146 @@
...
+
+        {/* Pages results */}
+        {results.pages.nbHits > 0 && !searchState.isFetching && (
+          <div className={searchPageStyle.searchResultsSection}>
+            <h3>Pages ({results.pages.nbHits})</h3>
+            {results.pages.hits.map(page => (
+              <div key={page.objectID} className={searchPageStyle.searchResult}>
+                <PublicationTitle>
+                  {/* FIXME this should be handled during indexing 
+                  (avoids external Link issues) */}
```
schwartzadev, line 125: This is tricky...it looks like _some_ `Link`s begin with '/', and others don't. We should probably standardize this across the codebase so we don't have to handle this here. Ideally someone could work some `grep` magic and find out where that's happening.

@kevee which is preferred?
1. prefixing `Link` with `/`
2. *not* prefixing `Link` with `/`

src/pages/search.js
superbiche: Totally agree. Once we've done the first review pass this ought to be a flexible component!

src/utilities/algolia.js
superbiche: Added some doc to get a better understanding of what it does and why.

src/pages/search.js
superbiche: Actually we *have* to prefix links with `/` otherwise Gatsby will throw warnings.
I'm pretty confident this is an inconsistency issue with the slugs and how these are built.

I could comment out this function to get a precise list of failing links, so we can check these individually

kevee:
This is going to be a feature in a larger project, so let's not merge it into master. I'll make a project branch for it to merge into tomorrow.

gatsby-config.js
```diff
@@ -20,6 +26,15 @@ module.exports = {
     'gatsby-plugin-eslint',
     'gatsby-plugin-remove-trailing-slashes',
     'gatsby-plugin-netlify',
+    {
```
kevee, line 29: I changed GATSBY_ALGOLIA_ADMIN_KEY to ALGOLIA_ADMIN_KEY. Variables prefixed with "GATSBY_" are automatically placed in a global variable on the client, and we don't want to share it.

gatsby-config.js
```diff
@@ -20,6 +26,15 @@ module.exports = {
...
+      resolve: `gatsby-plugin-algolia`,
+      options: {
```
kevee, line 31: Can we move the plugin definition to outside the config and make the entire plugin conditional on?

 `typeof process.env. ALGOLIA_ADMIN_KEY !== 'undefined' && (process.env.BRANCH === 'master' || process.env.CIRCLECI)`?

This means the algolia index will only get updated if the site is being built on our CI or in Netlify on master branch. The reason for this is we don't want to have to pass out the admin key to all developers, and after we go live any dev doing work on this is going to get their own key only if they are doing search work.

src/pages/search.js
kevee: Prefix them all with `/`

The pages here all have their slug defined in Contentful, which automatically creates the slug based on the title.

kevee:
I've also made a separate admin key for both Netlify and CircleCI scoped to just index management.

gatsby-config.js
superbiche: OK, knew that CRA worked this way too. Will change it

gatsby-config.js
superbiche: Of course. Do you know if it's best to retain the plugin's position or it doesn't matter?

superbiche:
Thanks for your feedback @kevee @schwartzadev. Fixes and adjustments coming later today!

jasonlcrane:
I added the autocomplete, using the Reach UI Combobox. I kept the form that Michel has been using on the page as well, for discussion's sake. I am not sure what the best approach is, but one thing I was thinking about was the possibility of ditching this autocomplete, and having the search results with snippets that Michel has displaying on the page go full screen on search (so essentially still autocomplete). There are some potential accessibility concerns there I'm sure that we would need to work around.

src/context/search-context.js
```diff
@@ -118,7 +118,7 @@ export async function querySearch(state, dispatch) {
   }
 }
 
-export function getSnippetResultOrExcerpt(hitType, hit) {
+export function getHighlightResultOrExcerpt(hitType, hit) {
```
superbiche, line 121: @jasonlcrane I think you were missing my last commit on this change, this may break the search results page.

src/pages/search.js
```diff
@@ -40,12 +52,87 @@ export default withSearch(({ search }) => {
     ? ` : ${query} (${totalHits} result${totalHits === 1 ? '' : 's'})`
     : ''
 
-  let searchEvent
+  const getSlug = item => {
+    if (!item.publishDate) {
```
superbiche, line 56: I think it's risky to rely on `item.publishDate`, imo adding a `type` argument to the arrow function would be safer

src/pages/search.js
```diff
@@ -40,12 +52,87 @@ export default withSearch(({ search }) => {
...
+      return `${item.slug}`
+    }
+    return `/blog/${item.slug}`
+  }
+
+  const goToResult = selectedItem => {
```
superbiche, line 62: It amazes me Combobox doesn't allow choosing what we pass. @kevee do you know if there's a way to pass the `item.objectID` instead?

src/pages/search.js
```diff
@@ -40,12 +52,87 @@ export default withSearch(({ search }) => {
...
+      ...results.blogPosts.hits,
+      ...results.pages.hits,
+    ]
+
+    const item = allResults.find(result => {
+      return result.name === selectedItem || result.title === selectedItem
+    })
+
+    if (item && typeof window !== 'undefined') {
+      window.location = getSlug(item)
```
superbiche, line 74: @jasonlcrane check out `withLocation` component, it passes `location` and `navigate` if I remember well.
You could add `navigate` to the component's props `export default withSearch(({ search, navigate }) => {` so you can then use it to go to the destination while retaining history.
(https://reach.tech/router/api/navigate)

src/pages/search.js
```diff
@@ -40,12 +52,87 @@ export default withSearch(({ search }) => {
...
+                No results found
+              </span>
+            )}
+          </ComboboxPopover>
+        ) : (
+          false
+        )}
+      </Combobox>
+      <form
+        style={{ marginTop: 50, paddingTop: 50, borderTop: '1px solid black' }}
```
superbiche, line 133: @jasonlcrane check my latest commits, I'm using the `.modules.scss` approach like elsewhere in the project and it's really cool

src/pages/search.js
```diff
@@ -80,7 +167,7 @@ export default withSearch(({ search }) => {
           <h3>Searching...</h3>
         </div>
       )}
-      <div className={searchPageStyle.searchResults}>
+      <div id="searchResults">
```
superbiche, line 170: @jasonlcrane this is also a regression against my last commits, could you fix it please?

schwartzadev:
image

- [ ] I don't think this is expected behavior...How does the autosearch prioritize results? I would expect the `Data xyz` pages to be the first results, not `Wyoming`, etc.

-----

image
- [ ] Design-wise, we should have a more clear indicator of what is a query and what is content (the query `covid` looks the same as `deprecated`, for example)

-----
image
- [ ] We should have spacing between the center bullet and the names/dates -- see the homepage for reference

----
Is there a reason for returning the whole page content under `Pages`? I would imagine something like the following would be more helpful:


Pages
----
Data API:
  * ...text surrounding the **_query_** from the user...
  * ...a **_query_** is provided by the user via the nav...

Another Page:
  * ...only one match for **_query_**.

-----
- [ ] It looks like the inline rendering is not working properly:
image
The sentence after `Javascript` should be on a new line, the text below it should not be indented.

-----
- [ ] `%` seems to be a wildcard (I'm getting all the states back, two pages, and both blog posts). If that's true, do we want to include that in production?

-----
- [ ] The accessibility statement renders as: `We welcome your feedback on the accessibility of COVID Tracking Project Website. Please let us` -- it looks like the end of the content is cut off. Maybe this has to do with the chunking processes?

src/pages/search.js
jasonlcrane: Yeah, you're right, that's brittle. The type is not known when `getSlug` is called, though. We could refactor it to loop over those hits arrays separately and set the type there, I think. That could get a little messy as well, but I'll see if I can come up with something that will work.

src/pages/search.js
jasonlcrane: Thanks for the reminder ... I was thinking this was temporary just while we play around with it, but my approach here should definitely be avoided.

src/pages/search.js
superbiche: Added `getSanitizedSlug` to generate slug depending on result type + handle missing starting `/`

gatsby-config.js
superbiche: @kevee Changed to `process.env.ALGOLIA_ADMIN_KEY`. Local build is passing.

gatsby-config.js
superbiche: @kevee fixed

src/pages/search.js
superbiche: Imo looping over the results while retaining the type is the way to go, added `searchResultTypes` const to make it easier and gonna make result keys use keys from this constant so it should be more robust

jasonlcrane: Used your new keys and `getSanitizedSlug` function to improve this.

src/pages/search.js
jasonlcrane: Thanks, good call. Updated.

src/pages/search.js
jasonlcrane: Updated to use the .module.scss file 👍

PR #703, opened 2020-04-24
Add some tests for state-nav.js component

gitjeff05:
## Description

Adding tests for and improving coverage for `src/components/pages/data/state-nav.js`. The coverage for this component was very low.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -20,6 +20,7 @@ module.exports = {
   },
   testURL: `http://localhost`,
   setupFiles: [`<rootDir>/loadershim.js`],
+  setupFilesAfterEnv: ['<rootDir>/setupTests.js'],
   collectCoverageFrom: [
     './src/**/*.js',
     '!./src/pages/**/*.js',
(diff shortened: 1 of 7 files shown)


jest.config.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   },
   testURL: `http://localhost`,
   setupFiles: [`<rootDir>/loadershim.js`],
+  setupFilesAfterEnv: ['<rootDir>/setupTests.js'],
```
gitjeff05, line 23: @kevee I changed this [per the enzyme docs](https://enzymejs.github.io/enzyme/docs/installation/).

src/__tests__/components/pages/state/screenshots.js
```diff
@@ -1,12 +1,10 @@
 import React from 'react'
 import renderer from 'react-test-renderer'
-import Enzyme, { shallow } from 'enzyme'
-import Adapter from 'enzyme-adapter-react-16'
```
gitjeff05, line 4: Because we now have a setup file, we don't need these in each test file where we want to use enzyme.

kevee: Oh, nice!

jest.config.js
kevee: 👍

PR #705, opened 2020-04-24
Add API testing to CircleCI

kevee:
## Description
This adds a new `api` workflow which runs the API build, and does Jest tests to make sure all the files were created, and that required fields exist.

The `build` workflow was renamed to the more aptly-named `gatsby-build`, and assets from the API (all the data downloaded) is now passed to build so it's doing a full production build off real data. The `gatsby-build` now depends on a successful API run, if the API workflow fails, so does build.

The JSON files will not output null fields, so we are just checking CSV file content for QA.

## Go-live time
Let's hold off on merging for Monday so I can prepare a writeup to the team about the impacts of this on dev work.


build/__tests__/post-build/v1/states/current.js
```diff
@@ -0,0 +1,53 @@
+const fs = require('fs-extra')
+const csv = require('csv-parser')
+const checkFile = require('../../../utitlities/check-file')
```
schwartzadev, line 3: Yay Typo CI!
`utitlities` --> `utilities`

build/__tests__/post-build/v1/states/all-states.js
```diff
@@ -0,0 +1,15 @@
+const fs = require('fs-extra')
+const checkFile = require('../../../utitlities/check-file')
+
+describe('File : US : All states', () => {
+  it('generated files for every state', async done => {
```
schwartzadev, line 5: Looks right to me 🎉

build/__tests__/post-build/index.js
```diff
@@ -0,0 +1,7 @@
+const fs = require('fs-extra')
+
+describe('Build has occured', () => {
```
schwartzadev, line 3: Typo CI again
`occured` --> `occurred`

build/__tests__/post-build/v1/states/current.js
```diff
@@ -0,0 +1,53 @@
...
+    checkFile('_data/v1/states/current')
+  })
+
+  it('matches field pattern', done => {
+    fs.createReadStream('_data/v1/states/current.csv')
+      .pipe(csv())
+      .on('data', row => {
+        expect(row).toEqual(
+          expect.objectContaining({
+            state: expect.anything(),
```
schwartzadev, line 16: I _think_ these checks will pass if we add new attributes to the states object...is there a way to flag fields that are being returned but are _not_ being checked with a warning?

(i.e. say we add an `antibodiesPositive` field: my guess is that these checks will pass, but won't be checking for the presence of that field. I'd love to see a warning that says the `antibodiesPositive` field isn't being checked in the tests.)

build/__tests__/post-build/v1/press.js
```diff
@@ -0,0 +1,38 @@
...
+    checkFile('_data/v1/press')
+  })
+
+  it('matches field pattern', done => {
+    fs.createReadStream('_data/v1/press.csv')
+      .pipe(csv())
+      .on('data', row => {
+        expect(row).toEqual(
+          expect.objectContaining({
+            title: expect.anything(),
```
schwartzadev, line 16: (blanket question) Is there any reason to use `expect.anything()` instead of more specific tests like `.toBe('string')`?

At the very least, we should probably validate things that could break down the line (urls, dates, numbers).

build/__tests__/post-build/v1/urls.js
```diff
@@ -0,0 +1,23 @@
...
+const checkFile = require('../../utitlities/check-file')
+
+describe('File : Urls', () => {
+  it('exists', () => {
+    checkFile('_data/v1/urls')
+  })
+
+  it('follows field patterns', async done => {
+    const results = await fs.readJSON('_data/v1/urls.json')
+    // To-do: Filter and StateID should be in every field row, but it is missing from a few states
```
schwartzadev, line 11: Just flagging this todo

build/__tests__/post-build/v1/states/current.js
kevee: HA!

build/__tests__/post-build/v1/press.js
kevee: Right now because the JSON files don't list values if there is nothing (i.e. if the grade is empty, then instead of `"grade":null` we get no grade value at all) we are using the CSV files for testing. It is a little janky, but it works.

However, CSV files turn everything into a string or null value.

build/__tests__/post-build/v1/urls.js
kevee: Yes!

build/__tests__/post-build/v1/states/current.js
kevee: We can't do that with built-in jest tools, but I'm going to add another expect statement that counts the properties.

schwartzadev: Smart -- thanks

build/__tests__/post-build/v1/press.js
schwartzadev: Ah, right. Hopefully we can be more comprehensive about that down the line, then.

Issue #457, opened 2020-04-09
Table redesign


kevee:
Meta issue of #426, #317, and #379, which have been closed and merged into this one:

We need a flexible way to handle tables across all devices. Let's create a new table component that is:

- **Responsive** - Stacks and reorients itself in mobile devices (#317 has lots of ideas)
- **Customizable** -  users can add/remove columns or column groups
- **Stable for linking** - Customizations by the user should be preserved for easy sharing/linking

This table component will first be applied to:

- [US Historical Data](https://covidtracking.com/data/us-daily)
- [State history views](https://covidtracking.com/data/state/california)

toddparker:
Ok gang let’s get a conversation on what we think is a good approach. In general, the most popular patterns for responsive tables are:

**Overflow** - just wrap it with a side scrolling container (current site does this). This is obviously the easiest approach but since the first cell scrolls away, you lose the context of what each row is so it can be disorientating and you need to scroll around a lot. Truly fixing the first column to keep context may still not be possible with CSS(?) but we may be able to do some hackery to keep it fixed (accessibility and layout robustness are key considerations) - https://github.com/COVID19Tracking/website/issues/317#issuecomment-607613708

**Stack** - re-format the table so each cell becomes a stacked ‘row’ and the column headers become labels. This is best for situations where comparing data between rows isn’t important because the stacked style means that values are physically far apart. The main advantage is everything is visible on-screen:  no need to scroll or toggle columns (content complexity is a consideration, lowt of rows and columns will stack out to a very long page)  - https://github.com/COVID19Tracking/website/issues/317#issuecomment-609918295

**Responsive columns** - prioritize the columns based on their relative importance and hide less important columns as the viewport shrinks so the table fits on-screen and ideally provide a way to let people override these defaults and pick which columns are most important to them. This mode keeps the context (usually the first cell with the name) and allows for easy row-by-row comparisons at the expense of seeing all the columns (discovery of the column picker is a key consideration) - https://github.com/COVID19Tracking/website/issues/317#issuecomment-610467846

**Collapsibles** - some of these ideas could be combined with expanding/collapsing content on demand. For example, we could show a simple table with just a few columns on a phone (or just the state name) and expand to show a stacked presentation of all the data values on demand. This can help manage the complexity of many columns of data presented in stack mode so you can understand the lay of the land - https://github.com/COVID19Tracking/website/issues/317#issuecomment-609414661

kevee:
[...]

Issue #716, opened 2020-04-26
Add timezone to screenshot times


schwartzadev:
Please let us know the page you are talking about, and your content suggestions!
![image](https://user-images.githubusercontent.com/18607205/80315319-fb1b7d80-87b3-11ea-869b-6633b7a08f9b.png)

See the title -- I would suggest adding a note to the historical tables like: `All screenshot times are in ET` instead of repeating the timezone all over the table. (Right now, we do neither of these.)

I do not know if the screenshots times are in ET or not -- the above is just an example.

kissane:
Fully agree with adding timezone (ET) and note, thank you @schwartzadev !

kevee:
They are in ET. How about just change the heading from Screenshot to **Screenshot (ET)** and then add a note in the footer of the table?

schwartzadev:
#728 will close this

PR #720, opened 2020-04-27
feat(tables)

### LINKED ISSUE #721: Tables: Improve typography and general styling
This is part of the larger table work in #457

For this first pass, let's create a branch and focus on taking a pass at improving the table styles for improved clarity. We should also look at our underlying table semantics to make sure these are as accessible as possible.

**Note**: In this first pass, we're not yet adding in responsive enhancements other than the simple overflow we have now. We will make issues for those features.

## Designs

Details for the design are in the **[Figma file](https://www.figma.com/file/fKUKsYMNw3NLSyIEOI1HKc/v2-refresh?node-id=95%3A2627)** but here are some highlights:

### Core table styles
- [ ] **Table headers** (`th`) are 13px heavy and should match the horizontal alignment of their corresponding cells (right if numeric, left if text)
- [ ] **Grouped column headers** are 17px heavy
- [ ] **Cells** are 16px regular weight and should be aligned right if they contain numbers, left if just regular text like the state name or date
- [ ] **Rows** should have a delimiter to help aid in horizontal scanning across wide tables. We're unsure of which approach we want to use so can we create _classes_ to switch between bottom row lines `1px solid rgba(0,0,0,.1)` or alternating (zebra) row fills `rgba(0,0,0,.03)`
- [ ] **Grouped columns** should have a left border on each TH and TD to chunk the columns for easier scanning. These should be darker than the row lines - `1px solid rgba(0,0,0,.2)`
- [ ] A **row hover** to add a bg color to the rows when you `:hover` or `:focus`. With the line dividers, the hover could be the zebra bg color. For zebra, it gets more complicated but maybe we can just increase the opacity of the RGBA fills %?

(Ignore sorting for now)

image

dasulit:
## Description
This PR improves typography and general styling of tables.

screenshot of table design

## Related Issues
Related to #457, fixes #721

### DIFF
--- a/src/components/common/table.js
+++ b/src/components/common/table.js
@@ -6,14 +6,14 @@ const Th = ({ children, alignLeft, scope, colSpan }) => (
   <th
     scope={scope}
     colSpan={colSpan}
-    className={alignLeft ? tableStyle.alignLeft : ''}
+    className={alignLeft && tableStyle.alignLeft}
   >
     {children}
   </th>
 )
 
 const Td = ({ children, alignLeft }) => (
-  <td className={alignLeft ? tableStyle.alignLeft : ''}>{children}</td>
+  <td className={alignLeft && tableStyle.alignLeft}>{children}</td>
 )
 
 const Table = ({ children, tableLabel }) => (
--- a/src/components/common/table.js
+++ b/src/components/common/table.js
@@ -2,6 +2,16 @@ import React from 'react'
 import DetailText from './detail-text'
 import tableStyle from './table.module.scss'
 
+const Th = ({ children, alignLeft }) => (
+  <th scope="col" className={alignLeft ? tableStyle.alignLeft : ''}>
+    {children}
+  </th>
+)
+
+const Td = ({ children, alignLeft }) => (
+  <td className={alignLeft ? tableStyle.alignLeft : ''}>{children}</td>
+)
+
 const Table = ({ children, tableLabel }) => (
   <div>
     <table className={tableStyle.table}>{children}</table>
@@ -9,4 +19,4 @@ const Table = ({ children, tableLabel }) => (
   </div>
 )
 
-export default Table
+export { Th, Td, Table }
--- a/src/components/common/table.js
+++ b/src/components/common/table.js
@@ -2,8 +2,12 @@ import React from 'react'
 import DetailText from './detail-text'
 import tableStyle from './table.module.scss'
 
-const Th = ({ children, alignLeft }) => (
-  <th scope="col" className={alignLeft ? tableStyle.alignLeft : ''}>
+const Th = ({ children, alignLeft, scope, colSpan }) => (
+  <th
+    scope={scope}
+    colSpan={colSpan}
+    className={alignLeft ? tableStyle.alignLeft : ''}
+  >
     {children}
   </th>
 )
--- a/src/components/common/table.module.scss
+++ b/src/components/common/table.module.scss
@@ -7,7 +7,7 @@
 .table {
   line-height: 1rem;
   border-collapse: collapse;
-  margin: 0.5rem 0 0;
+  margin: 0.5rem 0;
   font-variant-numeric: lining-nums tabular-nums;
   width: 100%;
 
@@ -48,6 +48,7 @@
       &[scope='col'] {
         text-align: right;
         font-size: 13px;
+        font-weight: 800;
       }
 
       &[scope='col'].align-left {
@@ -57,6 +58,7 @@
       &[scope='colgroup'] {
         text-align: left;
         font-size: 17px;
+        font-weight: 800;
         padding: 0.5rem;
       }
     }
--- a/src/components/common/table.module.scss
+++ b/src/components/common/table.module.scss
@@ -31,28 +31,28 @@
 
   thead {
     border-bottom: 1px solid $color-slate-400;
+  }
 
-    th {
-      padding: 0 0.5rem 0.5rem;
-      vertical-align: bottom;
+  th {
+    padding: 0 0.5rem 0.5rem;
+    vertical-align: bottom;
 
-      &[scope='col'] {
-        text-align: right;
-        font-size: 0.8rem;
-        font-weight: 800;
-      }
+    &[scope='col'] {
+      text-align: right;
+      font-size: 0.8rem;
+      font-weight: 800;
+    }
 
-      &[scope='col'].align-left {
-        text-align: left;
-      }
+    &[scope='col'].align-left {
+      text-align: left;
+    }
 
-      &[scope='colgroup'] {
-        text-align: left;
-        font-size: 1.05rem;
-        font-weight: 800;
-        padding: 0.5rem;
-        min-width: 10rem;
-      }
+    &[scope='colgroup'] {
+      text-align: left;
+      font-size: 1.05rem;
+      font-weight: 800;
+      padding: 0.5rem;
+      min-width: 10rem;
     }
   }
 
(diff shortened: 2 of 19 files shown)


src/components/common/table.module.scss
```diff
@@ -12,58 +17,65 @@
...
-    &[scope='colgroup'] {
-      font-size: 1.1em;
-      text-align: center;
-    }
-  }
+      &[scope='col'] {
+        text-align: right;
+        font-size: 0.8rem;
+        font-weight: 800;
+      }
```
beep, line 52: Just curious: could these two selectors be simplified to just the `th`? (I’d assumed all the `thead th` elements would be `scope="col"`.)

(Also, not a blocker for merge; really just curious!)

src/components/common/table.js
```diff
@@ -2,11 +2,25 @@ import React from 'react'
 import DetailText from './detail-text'
 import tableStyle from './table.module.scss'
 
+const Th = ({ children, alignLeft, scope, colSpan }) => (
+  <th
+    scope={scope}
+    colSpan={colSpan}
+    className={alignLeft ? tableStyle.alignLeft : ''}
```
beep, line 9: Is it possible to squelch this attribute entirely, if there’s no prop passed to the className? (Rather than rendering `class=""` in the markup.)

src/components/common/table.js
```diff
@@ -2,11 +2,25 @@ import React from 'react'
...
+  >
+    {children}
+  </th>
+)
+
+const Td = ({ children, alignLeft }) => (
+  <td className={alignLeft ? tableStyle.alignLeft : ''}>{children}</td>
```
beep, line 16: [Same question RE: empty class attributes](https://github.com/COVID19Tracking/website/pull/720/files#r418136719) as before!

toddparker:
@dasulit let me give you a recap so you know what to do for each of @beep's comments above because we had a healthy debate in Slack for each so we can get this to the finish line 🥇 

- [ ] **Totals table on /data** - The topmost table is using the same `<SummaryTable>` component as the states but this has fewer columns because the data can't be aggregated for some of these. That component accepts a prop that allow us to toggle the entire middle section. _We've decided to punt on this issue for now, we can improve it later_ [Slack thread](https://covid-tracking.slack.com/archives/C0118KE8L10/p1588274536312300)

- [ ] **Screenshot lists on state pages** - @beep opened a [pull request](https://github.com/dasulit/website/pull/1) for you to review and merge against this PR. It tweaks the screenshot link list alignment and removes the space between the time and am/pm. See [Slack thread](https://covid-tracking.slack.com/archives/C0118KE8L10/p1588274961316800)

- [ ] **Visual jog on the cumulative totals table** - @beep's [pull request](https://github.com/dasulit/website/pull/1) also addresses this by adjusting when the overflow kicks in on the table. See [Slack thread](https://covid-tracking.slack.com/archives/C0118KE8L10/p1588275253319600)

src/components/common/table.module.scss
dasulit: Good call! The additional level of nesting isn't needed, will fix.

kevee:
I just merged master into your branch because we made a change that was causing build tests to fail.

src/components/common/table.js
dasulit: Updated the ternary to a `&&`, the empty string should be squelched 👍

dasulit:
Thanks for the review @beep and summary @toddparker! I made @beep's suggested changes, and merged in the second PR. 😄 

The last commit here addresses the topmost `/data` table, would love a look at that fix. The left-right alignment might need another visual pass:

Screen Shot 2020-04-30 at 11 36 42 PM

I'm open to dropping that commit if we want to think more about the design first. Let me know what y'all think!

toddparker:
@dasulit (and @beep_ these changes look _fantastic_. I think this is ready for final review and mergin' 😍

src/components/pages/state/state-history.js
```diff
@@ -1,54 +1,58 @@
...
-        <th scope="col">Negative</th>
-        <th scope="col">Pending</th>
-        <th scope="col">Hospitalized</th>
-        <th scope="col">Deaths</th>
-        <th scope="col">Total</th>
+        <Th scope="col" alignLeft>
+          Date
+        </Th>
+        <Th scope="col" alignLeft>
+          Screenshot
```
kevee, line 15: Per #716, let's rename this to "Screenshots (ET)"

toddparker: Ooh, good call @kevee. In fact, we can make this even better by using the [abbreviation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) element to provide a clearer, more accessible title. Do you think we should use `EST` instead of `ET` - seems that's the [standard](https://www.timeanddate.com/time/zones/).

`(<abbr title="Eastern Standard Time">EST</abbr>)`

This gives you a tooltip and audibly reads that back on a screen reader

image

kevee: Yes, yes, and YES

tatianamac: My research on `abbr` is that it isn't widely supported based on recent testing (as in, the Screen readers (SRs) do not actually announce the `titles`:

- `abbr` isn't always announced with SRs as evidenced by [PowerMapper's tests](https://www.powermapper.com/tests/screen-readers/labelling/acronym-abbr-title/) and [Paciello Group's tests](https://thepaciellogroup.github.io/AT-browser-tests/test-files/abbr.html)
- SRs will often read it phonetically, which will [change the variation based on the word itself](https://accessibility.blog.gov.uk/2017/02/08/advice-for-creating-content-that-works-well-with-screen-readers/) (Scroll down to section about abbrs like VAT.

**Options**
- Use an `aria-label` in addition to `title`, replicating the same info. 
_This seems like the best option for this use case imho.
- Use a [glossary technique](https://stackoverflow.com/questions/45776312/how-do-screen-readers-read-abbr-tags).
_A future option for when it's acronym city_

kevee: I vote `aria-label`

careeningspace:
In viewing the new tables on my phone (android/chrome) the Recovered data/header disappeared. 
Screenshot_20200501-114413_Chrome

src/components/pages/state/state-history.js
toddparker: Thanks @tatianamac! I knew there would be some sort of curveball.  I agree that `aria-label` is a good idea. So we have _both_ that and `title`, like this so you still get a tooltip with a mouse:

(`<abbr title="Eastern Standard Time" aria-label="Eastern Standard Time">EST</abbr>`)

toddparker: @kevee are there other places across that site that have timezones states like this? Might be good to do a sweep and make them all consistent.

kevee:
Deployment todo (for me)

- [ ] Merge master into feature branch
- [ ] Lock publishing in Netlify
- [ ] Merge
- [ ] Test production deploy
- [ ] Turn auto-publishing on

src/components/pages/state/state-history.js
kevee: There are two others. I'll open a teensy PR to fix those, so don't worry about it in the context of the tables.

kevee: Just updated with a component for all timezones and updated the tables and other places that were using "ET"

beep:
I only did some drive-by CSSin’, @dasulit and @toddparker did _such heavy lifting_ here, amazing work all!!

toddparker:
Great work everyone, feels good to land our first PR.

PR #723, opened 2020-04-27
chore: Removed upper-case state and FIPS API paths

kevee:
## Description

In an effort to shrink the dev API snaphsots, and deploy file numbers, and because Netlify redirects paths to lower-case if they exist (like try https://covidtracking.com/DATA). Also the mixed upper-lower case paths were screwing up Windows builds. this PR:

1. Removes API paths `/api/v1/[upper-case]` i.e. `/api/v1/NY`
2. Removes all the FIPS state number paths in favor of defined redirects
3. Uses gatsby to generate redirects for all upper-case and fips urls to the new lower-case URLs.
4. Added redirects for static endpoints press, urls, screenshots
4. Adds the API fallback proxy to to covid.cape.io to the end


kevee:
Examples:

https://deploy-preview-723--upbeat-lovelace-3e9fff.netlify.app/api/v1/states/CA/info.json
https://deploy-preview-723--upbeat-lovelace-3e9fff.netlify.app/api/v1/states/03/info.json

PR #746, opened 2020-04-30
[WIP] Redirect API endpoints

kevee:
## Description

Based on a sample of 32,000 API calls being proxied to the old API endpoint at three different times of the day, this PR adds redirects to all possible used combinations of old API calls.

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -1,11 +1,14 @@
 /api/statesdaily state=:state date=:date      /api/v1/states/:state/:date.json
 /api/states state=:state                      /api/v1/states/:state/current.json
+/api/us date=:date                            /api/v1/us/:date.json
+/api/us/daily date=:date                      /api/v1/us/:date.json
 /api/v1/us.json                               /api/v1/us/current.json
 /api/v1/states.json                           /api/v1/states/current.json
 /api/statesinfo                               /api/v1/states/info.json
 /api/statesinfo.csv                           /api/v1/states/info.csv
 /api/statesdaily                              /api/v1/states/daily.json
 /api/statesdaily.csv                          /api/v1/states/daily.csv
+/api/states.json                              /api/v1/states/current.json
 /api/states                                   /api/v1/states/current.json
 /api/states.csv                               /api/v1/states/current.csv
 /api/us                                       /api/v1/us/current.json
--- a/static/_redirects
+++ b/static/_redirects
@@ -16,9 +16,10 @@
 /api/urls                                     /api/v1/urls.json
 /api/urls.csv                                 /api/v1/urls.csv
 /api/urls.html                                /api/v1/urls.html
-/api/screenshots                              /api/v1/screenshots.json
-/api/screenshots.csv                          /api/v1/screenshots.csv
-/api/screenshots.html                         /api/v1/screenshots.html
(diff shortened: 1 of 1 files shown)


kevee:
Sample endpoints:

https://deploy-preview-746--upbeat-lovelace-3e9fff.netlify.app/api/states?state=CA
https://deploy-preview-746--upbeat-lovelace-3e9fff.netlify.app/api/statesdaily?state=CA&date=20200401

webmasterkai:
Have you tried making a CORS request? I'm getting errors. You could try doing an internal proxy instead of a redirect.

You can remove any `.html` endpoints from the redirects. There is only CSV and JSON endpoints there.

As we have discussed many times, making these changes will break functionality for certain users that use query string combinations that are not accounted. For example `?grade=B` will not filter the list.

What is gained or improved by these changes?

webmasterkai:
Also, I remember the CSV files being used with query args. It doesn't look like there is provision for that in these redirects.

kevee:
Hi, Kai! So the redirects do support CORS. Here's the redirect header:
```shell
curl -I https://deploy-preview-746--upbeat-lovelace-3e9fff.netlify.app/api/statesdaily?state=CA&date=20200401
[1] 285
m043-300-28109:Documents mill5935a$ HTTP/2 301
access-control-allow-origin: *
```
and the v1 URL it redirects to:
```shell
curl -I https://deploy-preview-746--upbeat-lovelace-3e9fff.netlify.app/api/v1/states/daily.json
HTTP/2 200
accept-ranges: bytes
access-control-allow-origin: *
```

[Here's a codepen showing CORS working with the redirect and the old API URL](https://codepen.io/keveemiller/pen/JjYrPVv?editors=1010)

I agree that we want to not break anything as much as possible, but when I pulled the log of requests, I only saw `state` and `date` URL parameters, no like `grade=` entries. That prompted me to start this PR because it seemed like there weren't any edge cases that we weren't considering.

The idea behind them is to get all of our current API users hitting the same URLs, and to be able to track API usage through analytics. Proxies don't give us that ability in Netlify.

ccheever:
This looks good to me. Definitely makes things easier that only those 2 filters were being actually used.

webmasterkai:
I forgot about the analytics not only showing file views and not counting proxy requests. That's a valid point.

I've looked at some logs and of the top 2000 or so combinations. The only combination that isn't covered is folks using a combination of `state` and `dateChecked` field instead of `date` on /states/daily (`/states/daily?state=TX&dateChecked=`). Could we proxy that one combination? All other combinations not accounted in these redirects happen fewer than once every 8 hours. If someone complains they could be sent to the old covid.cape.io url and told it's available there but an "unsupported" option.

We will want to add some provision for CSV files
- `/states/daily.csv?state=DC&date=20200404`
- `/states/daily.csv?date=20200307`
- `/states/daily.csv?state=NY`

kevee:
Cool, thanks for the edits. We *do* see a lot of requests for `api/statesdaily` for some reason, so I think we should leave that redirect in there.

kevee:
⚠️ I am going to hold off on merging this PR until early AM. Also, Netlify is being super nice and is pulling all of their CDN data for us to take a look at.

PR #753, opened 2020-05-01
Update state searchbar to 100% for narrow screens

benhalpern:
#Description

Beauty is in the eye of the beholder, but I think this creates a more elegant mobile experience, where the search element is more clickable in 100% of the screen size.

It already takes up this space with empty space so I think this is a simple adjustment.

This is the new look on mobile:

Screen Shot 2020-05-01 at 10 20 22 AM

It's a pretty safe fix if you like it. Thanks for all the great work on this project!


benhalpern:
@beep nice thought, I added an adjustment to the PR. Thanks!

kevee:
I merged master in to make sure checks pass.

PR #768, opened 2020-05-04
feat: Add distinct field to search

### LINKED ISSUE #761: Search: chunked content return multiple results
Please let us know the page you are talking about, and your content suggestions!

Here's an example of the issue:
image

kevee:
## Description

This:

- Adds contentful_id to blog and page indexes
- Flags the field as distinct using `attributeForDistinct`, but doesn't filter them out in the index
- Uses `distinct` to only return one value per record

## Related Issues
Fixes #761

### DIFF
--- a/src/context/search-context.js
+++ b/src/context/search-context.js
@@ -111,9 +111,11 @@ export function useSearch() {
   return [useSearchState(), useSearchDispatch()]
 }
 
-async function queryIndex(index, query) {
+async function queryIndex(index, query, distinct = false) {
   try {
-    const hits = await index.search(query)
+    const hits = await index.search(query, {
+      distinct,
+    })
 
     return hits
   } catch (e) {
@@ -128,8 +130,8 @@ export async function querySearch(s, dispatch) {
   try {
     const [state, blogPost, page] = await Promise.all([
       queryIndex(stateIndex, s.query),
-      queryIndex(blogIndex, s.query),
-      queryIndex(pageIndex, s.query),
+      queryIndex(blogIndex, s.query, true),
+      queryIndex(pageIndex, s.query, true),
     ])
     NProgress.done()
     dispatch({ type: 'fetchSuccess', payload: { state, blogPost, page } })
--- a/src/utilities/algolia.js
+++ b/src/utilities/algolia.js
@@ -38,7 +38,7 @@ const blogPostQuery = `{
   posts: allContentfulBlogPost {
     edges {
       node {
-        objectID: contentful_id
+        contentful_id
         title
         authors {
           name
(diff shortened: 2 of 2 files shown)


src/utilities/algolia.js
```diff
@@ -196,8 +197,7 @@ const stateSettings = {
  * (see https://www.algolia.com/doc/guides/building-search-ui/going-further/backend-search/how-to/highlighting-snippeting/)
  */
 const pageSettings = {
-  attributeForDistinct: 'section',
-  distinct: true,
+  attributeForDistinct: 'contentful_id',
```
superbiche, line 200: @kevee I think this may lead to record dupes for the same page (https://www.algolia.com/doc/guides/sending-and-managing-data/prepare-your-data/how-to/indexing-long-documents/#after) as `contentful_id` would be the same across different chunks of a single page. If I got the docs right which is not sure at all.

src/context/search-context.js
```diff
@@ -111,9 +111,11 @@ export function useSearch() {
   return [useSearchState(), useSearchDispatch()]
 }
 
-async function queryIndex(index, query) {
+async function queryIndex(index, query, distinct = false) {
```
superbiche, line 114: Related to my first comment about the `attributeForDisctinct`: well done, what I understood from the docs was "either at indexing time or query time" but the "once attributeForDistinct is set, you can enable distinct" line just proved me wrong!

src/utilities/algolia.js
kevee: Right, I tested this and if you don't set `distinct` to true, then it should just say "use contentful_id when doing a distinct query" but the blog records all go into the index.

kevee: @superbiche - can you check the algolia test_ indexes and see if the blogs look right?

superbiche: @kevee looks good !

superbiche, line 200: Finally, it's not resolved: 
* https://www.algolia.com/apps/G1E1Q4HGTB/explorer/browse/live_blogPost
* Search `tracking race`
* We get 3 hits from the same page

The `section` attribute will filter these 3 results based on the auto-incremented value (section0, section1...). 

In fact, you may think the `contentful_id` is used to consolidate the result, like "all results with the same ID should be consolidated" but actually that's the opposite: we _want_ a different attribute for each chunk :)

kevee: @superbiche - I just re-updated the index, I think a local build had screwed it up:

https://deploy-preview-768--upbeat-lovelace-3e9fff.netlify.app/search?q=tracking%20race

Try "tracking race"

kevee: Also like Native American, which is in the second chunk for that post, shows up

superbiche: @kevee OK I totally understood the docs in the wrong way. The test index is OK, I guess you can change the live one to use  :+1:

kevee: Cool, thanks!

PR #788, opened 2020-05-05
chore: Use only lower-case paths in API

kevee:
## Description

- Removes upper-case state names in API routes, since Netlify automatically matches paths to lower-case.
- Removes unused FIPS-based file paths.


kevee:
Per note here, Netlify will serve anything as the lower-cased version:

https://community.netlify.com/t/my-url-paths-are-forced-into-lowercase/1659/2

PR #812, opened 2020-05-07
Refactor homepage chart design.

### LINKED ISSUE #807: change colors of the top for more contrast
- as per @jasonsantamaria's suggestion, @tealtan and I came up with slightly different colors for the top to solve the issue of contrast. 

background color: plum600 #575AAD
CTP bars: honey500 #E2894E
CDC bars: plum100 #F2F2FF 

Screen Shot 2020-05-06 at 5 02 04 PM

- we also decided on a new treatment for the numbers, so they don't look like links: 
Screen Shot 2020-05-06 at 4 45 22 PM

### LINKED ISSUE #808: spacing labels and font-sizes top
based on feedback we got from the design team, let's: 
- use font-size: 40px for main text and font-size: 18px for chart text 
- add a bit more spacing in the left side of the chart's y-axis labels, so they don't overlap with the chart on smaller sizes. this can also be solved by moving them to the left sooner 

axes-overlapped

cc @tealtan

### LINKED ISSUE #809: adjust height of top header
as per [discussion on Slack ](https://covid-tracking.slack.com/archives/C010N48RY73/p1588784465438100), can we adjust our elements to make the page look something like this? 

Screen Shot 2020-05-06 at 1 08 53 PM

I think we'll need to: 
- change header height and padding-top
- move the chart up a tad 

cc @tealtan

kevee:
## Description
- Darkens color in header and chart
- Redesigns color legends
- Resizes fonts
- Switches y axis to left when header stacks
- Brings chart higher up the hero element

## Related Issues
Fixes #807
Fixes #808
Fixes #809

### DIFF
--- a/src/components/layout/header.module.scss
+++ b/src/components/layout/header.module.scss
@@ -136,7 +136,7 @@
 }
 
 .container {
-  background: $color-plum-600;
+  background: $color-plum-550;
 }
 
 .site-header {
--- a/src/components/layout/header.module.scss
+++ b/src/components/layout/header.module.scss
@@ -1,14 +1,6 @@
 @import '../../scss/colors.module.scss';
 @import '../../scss/breakpoints.module.scss';
 
-.circles {
-  background: $color-plum-550;
-
-  .container {
-    background: none;
-  }
-}
-
 .tools {
   display: none;
   font-size: 0.81rem;
@@ -144,7 +136,7 @@
 }
 
 .container {
-  background: $color-plum-550;
+  background: $color-plum-600;
 }
 
 .site-header {
--- a/src/components/pages/homepage/header-hero.js
+++ b/src/components/pages/homepage/header-hero.js
@@ -192,9 +192,8 @@ export default () => {
             tests to date across the US. Using a rigorous data-collection
             process, we&apos;ve counted{' '}
             <span className={`${heroStyle.legend} ${heroStyle.ctp}`}>
-              {data.allCovidUs.nodes[0].posNeg.toLocaleString()}
+              {data.allCovidUs.nodes[0].posNeg.toLocaleString()}.
             </span>
-            .
           </p>
         </div>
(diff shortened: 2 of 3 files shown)


kevee:
@julialedur OK I removed all the top padding from the chart. Take a look once it's rebuilt!

julialedur:
so close!! i think making the chart a bit shorter should make the rest fit: 

Screen Shot 2020-05-07 at 12 09 23 PM

kevee:
Will do, in the meantime:

image

Should we move the period inside the legend?

image

sarawb:
I think period inside the legend will scan a lot nicer! (Love the new legend!)

kevee:
@julialedur how does the chart look to you now? Are we GTG?

kevee:
@sarawb Agreed! I moved the period.

tatianamac:
@kevee Not sure if it's too late but this colour combo of purple and orange fail colour contrast 😬

kevee:
Never too late for color contrast checking, @tatianamac! I was thinking it  looked a little dark. Here's my thoughts:

1. The color honey 400 `#ffad4a` passed WCAG AA for large text. That is what we currently have on the live site.
2. The color honey 500 `#e2894e` does not. That's what is in this PR.
3. *Text* color contrast is different than *chart* color contrast. Could we just keep the chart using 500, but change the label to 400? It would technically be different colors, but not by much. Only like um web designers would notice ;)

image

julialedur:
chart looks great, @kevee! thank you 🙏 

about the colors, i think we should avoid making the legend and bars different colors. the contrast issue was a tricky one, @tealtan and i spent some time trying to come up with a solution. i think once we make the numbers bold and adjust the spacing between bars, it will look better. 
i'm open to other oranges that contrast better if anyone has suggestions :)

kevee:
I updated the colors to a dark blue per @tealtan

Issue #764, opened 2020-05-01
Site Search: UI tweaks + cleanup/bugs


neeshabeee:
**Search box in Nav + Search Results page**

0) Accessibility improvements:
- On mobile, let's make the search field gain focus by default when the user has opened the menu. Basically, we want to remove the step of clicking into the search bar on mobile.
- Make the "Search" text to behave as a label. Remove the placeholder entirely. Below is a screenshot of what "Search" should now look like. The Figma file has also been updated to reflect this change, too.

![image](https://user-images.githubusercontent.com/22971446/81229604-f5d4e480-8fa4-11ea-9d5d-6c9d71197f4e.png)

1) When expanded, the left edge of the search box should align with the left edge of the text of the first listed nav item ("Our Data" in this case). Right now, the search box overshoots this a bit. I found that `width: 28.5rem;` seems to work great.

_This is what it currently looks like:_
![image](https://user-images.githubusercontent.com/22971446/80843819-0e668880-8bba-11ea-9cc9-b917e955c683.png)

2) Currently the X icon to clear a search is veryy close to the magnifying glass icon button - making it too easy to click the wrong thing. This is happening in both the nav + in the Search Results page. Adding 8px to the `padding-right` of the input area would do the trick, for both search boxes. 

_This is what it currently looks like:_
![image](https://user-images.githubusercontent.com/22971446/80844912-b4b38d80-8bbc-11ea-9d19-27820780dbb6.png)

3) Occasionally, I see that the magnifying glass icon with an unexpected color (gray when it should be blue and vice versa). I haven't yet found a way to reliably reproduce this - but will update this when I do. This also happens on the search box on the Search Results page.

_Here is a screenshot from when I got into this state:_
![image](https://user-images.githubusercontent.com/22971446/80842926-bb8bd180-8bb7-11ea-84d8-ff651b94b2e7.png)

4) Occasionally, I see the search box change shape as well. It should retain its rectangular shape.

![image](https://user-images.githubusercontent.com/22971446/80846522-62c13680-8bc1-11ea-8f2c-27e8dc7b1560.png) 

![image](https://user-images.githubusercontent.com/22971446/80846553-7a002400-8bc1-11ea-8b31-5cb4e94362c5.png)

**Autocomplete**

5) Can we hide duplicate options that appear in the autocomplete dropdown?
[...]

PR #822, opened 2020-05-08
Search UI & a11y fixes

superbiche:
##  Description
Fixes most issues from https://github.com/COVID19Tracking/website/issues/764
I can't test on iOS whether the search input changes shape. Would be very happy if someone could help me on that one.
Otherwise, all clear! 

Nothing index related was made so it's pretty safe to merge.

## Related Issues
https://github.com/COVID19Tracking/website/issues/764


kevee:
Thanks, @superbiche! It looks like some tests failed, I'm checking that out now.

beep:
Hey @superbiche, this is looking _so good_. Thanks for jumping on this, and so quickly.

I’ll do a proper review in a bit, but wanted to ask a question, and then quickly log two items I noticed.

Here’s **my question**! I noticed that the autocomplete field _only_ appears at the widest layout. Is there a reason it’s disabled for mobile/table users?

For the **two items I noticed**, here’s a quick screen recording I took of browsing the site with VoiceOver with macOS Safari:

https://www.dropbox.com/s/as4i8p3n8jt1g3s/voiceover%20-%20search.mov?dl=0

- The search panel is visually hidden by default on smaller resolutions, but you’ll notice that I can still browse to its contents with VoiceOver, or with my keyboard. If the panel’s not navigable via mouse/touch, it should be hidden with `display: none` until it’s opened. Otherwise, this leaves the user with a lot of material they need to tab through before accessing the logo and the menu button. (This also applies to [the “We just launched the COVID Racial Data Tracker” banner when the navigation panel’s opened](https://www.dropbox.com/s/1rj0w00zgd7v3a8/voiceover%20-%20search%20expanded.mov?dl=0).) This is a pretty important usability fix; accessibility-wise, I’d want us to consider this a blocker.
- _Very small style note_: the “Get Involved” button looks like it has a decent amount of spacing above it at the smallest resolution, but loses it around the midrange.

I think that’s it for now! Like I said, I’ll take a proper look later, but I think this is looking really great, @superbiche. Thanks again for tackling all these improvements, and so darned quickly.

superbiche:
@kevee my bad, forgot to generate the snaps (again).

@beep regarding your questions: 
1. If I remember well this decision was taken while coding it, but it may be related to a shortest path towards shipping the feature. I thing @kevee will know more. Also, if we want to keep it on mobile I can work on it.
2. I couldn't agree more. I don't test with VoiceOver (yet) but I do use keyboard navigation a lot. And on the a11y side it's clearly a blocker. Will fix this (probably today)
3. Great call for the Get Involved button. I'll check what could possibly happen to it!
4. Your clip shows some kind of strange behavior. When the menu is expanded, the "Search" label seems to be missing, but it's not: focusing it in the "background" made the `autocompleteHasFocus` event become `true`, then expanding the menu seem to toggle it in the wrong direction. I'll check if this still happens when the closed menu can't be tabindexed anymore.
 
I'll be working on this during the next hours! Thanks _a lot_ for the feedback!

beep:
> If I remember well this decision was taken while coding it, but it may be related to a shortest path towards shipping the feature. I thing @kevee will know more. Also, if we want to keep it on mobile I can work on it.

That’s real helpful Michal, thanks! And it makes perfect sense, especially given the timelines y’all were working under. The reason I asked the question is because generally—_generally!_—my perspective is if we’re delivering a feature for one set of users, we should make sure it’s available for everyone.

That said, I wouldn’t think that adding autocomplete on mobile should be a blocker for getting this PR out the door! The a11y fixes are much more important, imo. I’d say this is probably worth logging, and coming back to as we can.

> Your clip shows some kind of strange behavior. When the menu is expanded, the "Search" label seems to be missing, but it's not

Ah _nice_, @superbiche—I saw that, but couldn’t reproduce it. Thanks so much for figuring out what was happening!

kevee:
@beep yes, we decided not to ship with autocomplete because the experience was a little wonky with the fly-up menu and the fly-down autocomplete, which could cover the button to close the menu. I'll create an issue to track this.

beep:
_Tooooootally_ makes sense, @kevee! And sorry, I realize I sounded critical before! To be clear, I think the decision to get this out the door for widescreen was super sound, especially given the timelines. If we can tackle broader access for this at some point, that’ll be amazing!

kevee:
Oh and @superbiche I think the test errors being thrown aren't snapshots, but `searchInputRef.current.blur()` not existing in a headless browser. You might need to check that the ref exists before calling it.

kevee:
No criticism taken!

superbiche:
@kevee @beep I think we're good on both tests and avoiding rendering the mobile menu if it's not expanded.
@beep not rendering the mobile menu seems to work great, keyboard navigation feels great. Also this mechanically fixed the 4th issue like expected (the opposite focus styles).
And last, the Get Involved button margin is consistent!

superbiche:
I'm implementing the mobile autocomplete, so we can review it a11y wise and deploy everything at once

superbiche:
Haha ! _behold the mobile search autocomplete_ ! With flexible menu!
Let me know it this looks good.

src/components/layout/header.js
```diff
@@ -82,35 +82,52 @@ const HeaderSearch = ({ children }) => {
...
+      setMenuHeight({ ...menuHeight, current: menuHeight.initial })
+    }
+  }, [query, isFetching])
 
   return (
-    <div className={headerStyle.mobileMenu}>
+    <div
+      ref={menuRef}
+      className={headerStyle.mobileMenu}
+      style={{
```
kevee, line 124: Now that we're applying show/hide of the menu using styles directly, instead of classes, it means if you tap the search box then resize, the mobile menu gets stuck open. Could we fix this by either applying an event listener to `onresize`, or even better instead of doing it via style, removing the `display` part of the style definition, and add/remove classes like:

className=`${headerStyle.mobileMenu} ${expanded === true && headerStyle.mobileMenuExpanded}`

image

superbiche: Good call. Will be fixed tomorrow!

superbiche: Actually had to do both:
* Add proper styling to avoid `style` taking precedence  
* Add a (throttled every 250ms) `resize` listener to actually "un-expand" the menu, so when you're in small res, open the menu, expand the page, and get back to small res, the menu isn't `expanded` anymore.

superbiche:
@kevee I think we're good now. Also a weird mobile search input width issue due to Combobox is now fixed.
I'm not @-ing Ethan as I learned he's back to client's work, but his first issue is fixed (no more rendering of the partnership banner when the mobile menu is expanded, so it's cannot be focused) and the second issue raised was related to your's.

PR #824, opened 2020-05-08
fix(tables): Improve table markup for accessibility

### LINKED ISSUE #724: Tables: Semantics for better accessibility
This is part of the larger table work in #457

I'd like to review our current table semantics and make sure we're following best practices:
- [WAI tables tutotrial](https://www.w3.org/WAI/tutorials/tables/)
- [Deque tables checklist for WCAG](https://dequeuniversity.com/checklists/web/tables)

## To review
We should pay special attention to grouped or spanned columns/rows

- [ ] [US Totals History](https://covidtracking.com/data/us-daily) (flat)
- [ ] [State history](https://covidtracking.com/data/state/alabama#historical) (flat)
- [ ] [State summary](https://covidtracking.com/data/state/new-york) (grouped)
- [ ] [Totals by state](https://covidtracking.com/data) (grouped)

dasulit:
## Description
Cleans up table markup for improved accessibility. 

## Related Issues
Fixes #724.


dasulit:
Hey @toddparker @beep @tatianamac! Leaving some notes on the nocturnal work-in-progress here, in case you all get a chance to look tomorrow during the day.

- I've focused on getting the header markup correct. This should be in place for the `summary-table` component, next up is making sure the other tables get the same attention.
- Known styling TODOs: Need to re-add in the left group border, need to address some funkiness with header width variability:
Screen Shot 2020-05-07 at 11 50 41 PM

* I added a Reacty `renderColumnHeaders` to make maintaining the markup for headers/columns a bit easier, but let me know if it's preferred to maintain the columns manually.
* `role="text"` is setting off an eslint warning:

```
Elements with ARIA roles must use a valid, non-abstract ARIA role.
```

I've commented out the `jsx-a11y/aria-role` rule for now, but wondering if it makes sense to choose a different role here? Perhaps `role="rowheader"`?

beep:
Thanks, @dasulit! So so exciting to see this take shape. I’ll try to take a proper look later, but wanted to quickly weigh in on this point:

> `role="text"` is setting off an eslint warning:

As you and eslint have noticed, `role="text"` is an invalid ARIA role. (A bit of history here, if you’re at all interested](https://incl.ca/roletext-presently-kinda-not-thing-sorta/).) That said—and this is thoroughly gross!—it’s needed _in this particular case_ to [prevent VoiceOver from “splitting” the words](https://axesslab.com/text-splitting/).

I wouldn’t suggest disabling the rule globally, as invalid ARIA roles _should_ be discouraged. But if there’s a way to add a separate test (or an exception) for this particular component, that might be the best way forward.

toddparker:
Great progress on this @dasulit! I think we landed with adding a class on the first `td` in each group to add that left border so hopefully you can borrow that code from @beep's sample.

One small thing I noticed (Safari/Mac) around 11px viewport, [the On Ventilator](https://deploy-preview-824--upbeat-lovelace-3e9fff.netlify.app/data/state/alabama) header looks right aligned and the Total Test Results looks center aligned (should all be left). It's ok on smaller/bigger viewports, odd.
image

I know you had logic to left align the first header in a group, maybe that regressed? Seems that both have right aligned text:

image


image

tatianamac:
When including that special exception @beep so eloquently suggested, would it make sense for us to write some comments in the code to future proof it from a someone who may come across it and think it was an unintentional error?

(I defer to those more familiar with the code base!)

dasulit:
Thanks team, this should be most of the way there now! Markup should hopefully be improved, and styling preserved. I've also added a comment about the `role="text"` exception.

One styling bug I have not yet been able to resolve: I offset the group header from the _top_ instead of the _bottom,_ as in @beep's example, to ensure the correct spacing when the group header is two lines long. Alas, this then doesn't work out when the column header is two lines long, which happens on narrower screens.  

Screen Shot 2020-05-11 at 9 56 59 AM

I think this can be fixed by offsetting the group header relative to the column header, but I haven't solved this yet. Will be able to tinker again tonight, but open to someone else giving it a shot if they have free hands today!

---

Otherwise, a couple other minor style differences I'd like feedback on:

I've left aligned the summary table at the top of the data page, this felt better after looking at it for a while. Let me know what you think.
Screen Shot 2020-05-11 at 9 47 29 AM

Longer group headers wrap with the column header beneath. See how `On Ventilator` takes up two lines instead of one. Perhaps we could set an arbitrary minimum width?
Screen Shot 2020-05-11 at 9 49 17 AM

cc @beep @toddparker @tatianamac

toddparker:
Thanks @dasulit this looks great! I'll work with @beep on a suggested alignment and min-width or something to control the wrapping. One technical feasibility question:

Can we set a width on the grouping header based on the count of columns it "spans" in React-land? ex. If there are 2 or less, add a certain class to the grouping header; if 3 add a different class, if 4 etc.

kevee:
@dasulit Thanks so much for all this work! Some thoughts:

I've run into that JSXa11y issue before, and here's a hack that the [jsx-a11y folks suggest](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-role.md#succeed). I'm doing an issue to see if they can fix it for role="text", or if not we can always patch that later.

```jsx
const role = 'text'

return (
<p role={role}>This is fine</p>
)
```

@toddparker's suggestion on applying a class based on colgroups can be done within `renderColumnHeaders` by [counting the columns in a group](https://github.com/COVID19Tracking/website/pull/824/files#diff-b5aa78b6c965524746a75b509e87be68R14) and applying a class.

toddparker:
Thanks @kevee! I amended my comment with some suggest values for min widths if @dasulit wants to tinker with that idea. Ultimately, the headings may not fit so we need to decide if we truncate to one line or handle wrapping.

Seems it would be ideal if the could wrap to 2 lines

@dasulit ok, so to recap:
- [ ] Set a `min-width: 5em` for all THs
- [ ] Set a `min-width: 10em` for the groupings text by default but use the column count to increase this if there are 3+
- [ ] Try tinkering with using a `width` instead of a `min-width` for the grouping text, that would. force the text to wrap if needed to it doesn't run over the next grouping with long text
- [ ] Let's let the heading wrap but ideally they would be bottom aligned so in most cases the heading would be right next to the THs and if it wraps it would go up. If this looks wonky, we could ellipsis the heading text, seem this won't kick in with any of the headings we have now anyway so this is just for safety

kevee:
OK, @dasulit I pushed a few changes:

- Added a `columnWidth` prop to `Th` that sets a class if it exists. These classes increment by `3em` for every column after 2. Nice thing about SCSS modules is only the ones we use are built in the site!
- Refactored `thClasses` variable to be an array we toss classes in and join with a space at the end.
- Added a `wide` prop to `Th` components that sets another class to make it `13em` - @toddparker even with `10em` it was too narrow for the last header to prevent overlap.

dasulit:
Thanks for those changes @kevee, they look great!

I made one small fix to return `null` rather than empty `<Td>` components for the main overview table, and removed the `eslint-disable` exception per your suggestion.

kevee:
Looks good, @dasulit. @toddparker are we ready to merge this?

toddparker:
I just did a QA pass on a bunch of browsers and this looks great. These min widths look good with the data we have so far so I think we're ready to go! Thanks for getting this across the finish line @dasulit and @kevee! Looking forward to having more accessible grouped headers.

toddparker:
https://bukk.it/celebration.gif

Issue #826, opened 2020-05-08
Search test coverage


superbiche:
**Describe the bug**
Follow up to @schwartzadev who brought up the fact we didn't write tests for the search feature :eyes: 

> I particular, I'm looking at these files:
context/search-context
utilities/algolia
components/layout/search-autocomplete

Assigning myself as I never wrote Jest tests, but _need_ to learn how to do this. If it's urgent and I'm not doing it feel free to participate.

PR #829, opened 2020-05-08
test: increase coverage by 16%

schwartzadev:
## Description
This adds a ton of Jest tests, mainly for blog and homepage components.

This does not create tests for search, which seems to be the majority of uncovered code at the moment. (See #826.)

## Related Issues
#826


schwartzadev:
With all of our new components, the % change in coverage dropped from 16% to ~7% (!!)

Issue #817, opened 2020-05-07
Create new Volunteer Application form


amandafrench:
Related to #816. As per discussions with @sarawb and @kevee, we need a Volunteer Application form to replace our current Google form. Please ping me privately for the group email address the form should be sent to. 

```
**Volunteer for the COVID Tracking Project**
Timestamp*
Name* [short text]

Email* [short text]
_If possible, this should be a Gmail or Google-linked address, since we rely heavily on Google Docs and Sheets. We will show your email internally to other volunteers but will not share it elsewhere._

URL [short text]
Personal website, LinkedIn, or other website that will tell us about you

About how many hours are you available to work?* [dropdown]
- 1-5
- 5-10
- 10-20
- 20-30
- 30-40
- 40+

What time zone are you in?* [dropdown]
- Pacific
- Mountain
- Central
- Eastern
- Other

When are you most available to work?* [checkboxes]
- Morning, some or all weekdays
- Afternoon, some or all weekdays
- Evening, some or all weekdays
- Weekends

Specializations / Skills* [short text]
_Examples: Python, SQL, Tableau, data viz, editing, social media, public health, research, journalism, etc._

What would you like to volunteer to do for us?* [radio buttons]
- I want to help run the whole program by creating high-level policies and services. 
- I want to recruit, train, and manage volunteers. 
- I want to collect COVID data from state websites and data dashboards. 
- I want to collect data about how COVID is affecting communities of color.  
- I want to maintain ratings of the completeness and quality of state COVID data. 
- I want to find COVID data by watching press conferences and monitoring local news sources. 
- I want to contact state governments and other entities for more and better data. 
- I want to write and edit copy for your website. 
- I want to create data visualizations and graphics with your data. 
- I want to design the front end of your website. 
- I want to build the back end functionality of your website. 
- I want to ensure your heavily-trafficked website is fast and reliable. 

I was referred by … [short text]

Why are you interested in volunteering for us? [long text]

kevee:
@friej715 is taking this on later today, here's some notes:

- This should be it's own page under `contact/volunteer`
- Include the Contentful snippet `volunteer-form` on top
- This needs to be a custom form, not use the "contact us" form component 
- Form name should be "volunteer"
- Make sure to wrap the radios with a `<fieldset>` and `<legend>` for a11y

amandafrench:
[...]

PR #831, opened 2020-05-09
feat(contact): first pass at volunteer page for #817

friej715:
## Description

Adding contact page for volunteers.

## Related Issues

Addresses #817. Also moves contact components to their own `contact` folder, so as not to clog up `common`.

## Notes:

Just a few things to clarify before going forward. My brain is melting from doing this after dayjob, so forgive me.

* Will the timestamp in the original comment on #817 be generated by Netlify?
* Required form markers (e.g. asterisks) seem to throw a bunch of wrenches in screen readers. I just went with the small "(required)" text in red, since that seems to be the best of all worlds. Is that okay?
* Is there a way to make a fieldset required without too much hassle?


kevee:
Thanks, @friej715! Some answers:

- **Timestamp** - Yup!, don't worry about this one
- **Required form markers** - I think it shoud just be a small (required) text within the `<label>` element. Also, add an `aria-required` attribute to the required input elements.
- **Making field required** - Let's just add a `required` attribute to the field, and if down the road we are getting a lot of submissions that are incomplete, we'll see if we can't change that.

friej715:
Copy that @kevee, on it after work! (Not sure why I didn’t see your comment earlier, I feel like I’m not getting emailed github notifications anymore :/ )

kevee:
Closing in favor of #926

PR #844, opened 2020-05-12
feature: Sub navigation

kevee:
## Description
Adds a sub-menu system for navigation


kevee:
A11y todo:

- [ ] Remove the `menu` role from `<Menu>`
- [ ] Add `tab` support within menu

kevee:
Go live:

- [ ] Remove contentful environment for this branch
- [ ] Lock publishing
- [ ] Remove top-level items in navigation in Contentful
- [ ] Merge
- [ ] Test
- [ ] Unlock publishing

tealtan:
Some notes!

The caret:
- [ ] I’d be curious if we can increase the tap target further to 40px by 40px – this is larger than even the circular button style used for touch, which is a good size currently. Basically we want to increase the affordance of a finger tapping and hitting the button, even though visually it will look smaller.
- [ ] The caret should hang vertically in the middle of the nav item.
  - On mouse: it’s a bit too low, should be higher by ~4px
  - On touch: it’s too high, should be lower by ~16px
- [ ] The focus outline color should be white

The "current" page style:
- [ ] I think I’m going to advocate for removing this style, since the current style (of changing opacity) and the previous style (of making it blue) take us towards having less contrast. I’d also be open to a simple underline, but not sure we really need it.

Updated hover styles:
- [ ] This could wait until after launch, but would be great to apply the new hover styles to all the things in the nav. For the top level items, that means the link should stay white and have a white underline on hover (removing the blue bottom border style)

Tabbing:
- [ ] Tabbing should work to navigate between dropdown items when expanded. I think they are out of the tab list because their tab-index is set to -1 – maybe this gets removed when the dropdown is expanded?

kevee:
The hover buttons are not round in iOS on Safari it Chrome.image

jasonlcrane:
I fixed that rounded button issue in Safari. Turns out there is a default padding that we had to reset. 
Screen Shot 2020-05-14 at 9 25 46 AM

PR #878, opened 2020-05-19
chore: Change data source to API repo

kevee:
## Description
This adds the real API endpoint to the race project dashboard. Notes:

- It uses data from `./_api` which is a cloned repo
- You must run `npm run setup` or to not wipe your envvars, `npm run setup:api-repo` to clone the data
- Some fields (duplicates) have changed, I updated the big GraphQL query, but didn't do QA on everything

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,9 +149,9 @@
     "test:lint": "eslint ./src",
     "test:dev": "run-s test:lint test",
     "test:api-post-build": "jest  --config=\"{}\" --testPathPattern=./build/__tests__/post-build/",
-    "setup": "run-s setup:api-data setup:env-vars",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -151,6 +151,7 @@
     "test:api-post-build": "jest  --config=\"{}\" --testPathPattern=./build/__tests__/post-build/",
     "setup": "run-s setup:api-data setup:env-vars",
```
schwartzadev, line 152: should this include `setup:api-repo`?

schwartzadev:
It looks like two things changed:
1. Notes are now coming through as empty strings
2. Something with the `Disp` flags is mixed up

image
i.e. for `Black or African American alone`, this is _not_ a disparity (6% = 6%), but this PR makes it seem like one.

PR #890, opened 2020-05-20
fix(logic): added language for death rate ratio cases

goldfarb:
## Description

[Added cases](https://github.com/COVID19Tracking/website/projects/10#card-38544863) for the deaths / percent of the population ratio, so that we get standard language instead of a specific number

image

* For deaths / population < 1.36, the sentence does not show up at all

* currently, the cases account for up to deaths / population of 3.35, for anything higher I included 'much higher'

### DIFF
--- a/src/components/pages/race/multiplier-highlight.js
+++ b/src/components/pages/race/multiplier-highlight.js
@@ -2,13 +2,53 @@ import React from 'react'
 import Container from '~components/common/container'
 import multiplierHighlightStyle from './multiplier-highlight.module.scss'
 
+function range(multiplier) {
+  if (multiplier >= 136 && multiplier <= 149) {
+    return 'nearly 1.5 times'
+  }
+  if (multiplier > 149 && multiplier < 151) {
+    return '1.5 times'
+  }
+  if (multiplier >= 151 && multiplier <= 185) {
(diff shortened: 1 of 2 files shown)


src/components/pages/race/multiplier-highlight.js
```diff
@@ -2,13 +2,53 @@ import React from 'react'
 import Container from '~components/common/container'
 import multiplierHighlightStyle from './multiplier-highlight.module.scss'
 
+function range(multiplier) {
```
kevee, line 5: Can we simplify this a little to:

```javascript
// letting multiplier be a float:
const divisor = multiplier % 1
if(divisor % 1 > 0.36 && divisor < 0.49) {
  return `nearly ${Math.floor(multiplier) + 0.5} times`
}
if(divisor % 1 >= 0.49 && divisor < 0.51) {
  return `${Math.floor(multiplier) + 0.5} times`
}
if(divisor % 1 >= 0.51 && divisor < 0.85) {
  return `more than ${Math.floor(multiplier) + 0.5} times`
}
if(divisor % 1 >= 0.85 && divisor < 0.99) {
  return `nearly ${Math.floor(multiplier) + 1} times`
}
...etc
```

PR #908, opened 2020-05-21
feat(content): Add a component to enforce smart quotes

### LINKED ISSUE #861: Add smart quote support to all content pages
Let's get some [smart quoting](https://smartquotesforsmartpeople.com/) into our content. How about we:

- [ ] Install `gatsby-remark-smartypants`, this should fix all MD stuff from Contentful
- [ ] Implement a filter for blog posts, since they use rich text format.
- [ ] Create a common smartQuote component that uses [smartypants](https://www.npmjs.com/package/smartypants) package (see below)
- [ ] Manually audit all static content pages and wrap text with this component:
  - [ ] Visualization guide
  - [ ] Homepage
  - [ ] Racial data page

Sample idea, but this is flimsy and would need to make sure it's only being passed text elements:

```javascript
import React from 'react'
import smartypants from 'smartypants'

export default ({children}) => (
  <>{smartypants(children, 1)}</>
)
```

dasulit:
## Description

Add a component to enforce smart quotes.

## Related Issues
Fixes #861

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16990,16 +16990,33 @@
       }
     },
     "gatsby-remark-smartypants": {
-      "version": "2.2.3",
-      "resolved": "https://registry.npmjs.org/gatsby-remark-smartypants/-/gatsby-remark-smartypants-2.2.3.tgz",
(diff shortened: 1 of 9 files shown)


kevee:
Hi, Dana! Looks like there's a merge conflict, but we can't get a preview site until you merge master in with your branch.

PR #943, opened 2020-05-28
chore(docs): Update availability language on volunteer form etc.

amandafrench:
Description

Update availability language on volunteer form, point to new success page, fix field labels 

Related to #817 

Post-merge will need to publish new page titled "Thank you for wanting to volunteer" at /contact/volunteer-success


amandafrench:
Oh, okay! Sorry, I can change them back. Was trying to fix that weird thing where the form output on "availability" and "workstream" is using the first option as the label: 

Screenshot 2020-05-28 07 45 19

Screenshot 2020-05-28 07 46 25

PR #949, opened 2020-05-29
Handle no-js for CRDT Dashboard Key

### LINKED ISSUE #911: "not comparable" / "disparity likely" dialogs inaccessible without Javascript
**Describe the bug**

## Current behavior
I hate for Javascript to be My Thing, _but_... There is currently no way to view these dialogs without using Javascript.

image
(Open links are highlighted)

### Reproduction
1. Disable JS
2. Go to https://covidtracking.com/race/dashboard
3. Click on either `See why ↓` span

## Suggested behavior (1)
These links (`See why ↓`) could link to a `#key` element at the bottom of the page, with the un-collapsed notes.

We could also pair this with a link at the top/in the hero about "what does 🔵 and 🚩 mean?"

### Drawbacks
There wouldn't be a great way to return to the state from which you jumped to the bottom of the page

## Suggested behavior (2)
Both dialog could be displayed below the key, for every state.

### Drawbacks
This would be super repetitive.

schwartzadev:
## Description
Makes the table key descriptions accessible without JS.
Fixes #911


kevee:
Good idea just making them links to a footnote! Query: there is a weird formatting error that erases the key and makes the first key bigger when JS is disabled. Ideas?

image

schwartzadev:
Good catch, @kevee. I'm in a meeting now, but I'll check it out soon. It looks like something is wrong with the `<p>` tags ... I'll probably refactor the key elements into components.

schwartzadev:
Hrm so I'm still seeing the same issue with JS: the DOM is different with and without Javascript...

Here's what I'm seeing with JS: :heavy_check_mark: 
image

...and without: :x:
image

It looks like the content of the `TableKeyContentItem` component is not rendering inside the `p` tag...

I'm not exactly sure how to fix this...could this be something wrong with Gatbsy?

kevee:
@schwartzadev I was doing a PR for cleaning up note sources, so I'm going to address the no-JS issue as well in that one and close this.

The problem was that the inner element was a `div` inside of a `span`, which is verboden. I'll see if I can't catch that with ESlint.

PR #951, opened 2020-05-29
chore(docs): Tweaks volunteer form final question

amandafrench:
## Description

Related to #817 

Changes final question to "What kinds of projects or tasks would you most like to help with?"


amandafrench:
Green to go, @kevee, thanks -- in the future should I be editing the snapshots directly in addition to the form?

kevee:
@amandafrench nope, just run `npm run test:update` and it should happen automatically.

PR #954, opened 2020-05-31
Add CRDT social card

### LINKED ISSUE #888: Enable UNFURL when sharing page(s) on social media
Would be cool to unfurl website links, so we have some visual to accompany direct links. This is especially evident, given the new white paper. May need to open another ticket to ensure proper graphic is ready for unfurling.

schwartzadev:
## Description

Closes #888: adds a custom social card to CRDT pages instead of the default card.

@kevee I'm leaving this as a draft for now since I'd love to add a flag to the contentful pages (sourced from `allContentfulPage`) that would support this behavior. This would require updating the contentful model to support a `isRace` flag, with a default value of `false`. Provided you're on board with this change, I can make it. The current use case is at [covidtracking.com/race/about](https://covidtracking.com/race/about)


schwartzadev:
It looks like these changes are working, according to [Twitter](https://cards-dev.twitter.com/validator) and Slack:
image

kevee:
@schwartzadev I like this PR, but your comment gave me this idea. How about we merge this for now then I'll open a new issue to manage social cards *within* Contentful as a new content type. That way pages in Contentful can just be related to cards, or pages in the codebase can reference card slugs like how we manage snippet content now.

kevee:
Added #955

schwartzadev:
Perfect, #955 makes tons of sense. I'll merge this for now and start on #955.

PR #957, opened 2020-06-01
Add contentful social cards

### LINKED ISSUE #955: Social cards in Contentful
Let's manage our social cards in Contentful.

- [x] Add social card content type to Contentful with name, slug, image, and description
- [x] Add field to pages in Contentful that relates to social cards
- [x] Let the  `SEO` components accept a `socialCard` prop type that represents a card object.
- [x] The `Layout` component should have `socialCard` that it passes to SEO
- [x] The `src/templates/content.js` template should query for socialCard of pages and pass the card through that prop
- [x] Handle social cards from blog posts
- [x] Pages like `src/pages/race.js` can query for that card directly, just like we do with snippets, and pass the card object to `socialCard`
- [x] Remove the older `description` prop from `Layout` and it's use in a few pages along the site.
- [x] Replace pages that used description with references to social cards in Contentful
- [x] Test with both [FB's sharing debugger](https://developers.facebook.com/tools/debug/) and the [Twitter card validator](https://twitter.com/login?redirect_after_login=https%3A%2F%2Fcards-dev.twitter.com%2Fvalidator)

schwartzadev:
## Description
Closes #955: `Let's manage our social cards in Contentful.`


src/components/utils/seo.js
```diff
@@ -2,24 +2,42 @@ import React from 'react'
...
 
-  const card = isRace ? crdtCard : defaultCard
+  const defaultSocialCard = contentfulSocialCard
+  let description
+  let imageSrc
+  if (!socialCard) {
+    description = defaultSocialCard.description.description
+    imageSrc = defaultSocialCard.image.resize.src
+  } else if (socialCard.description && !socialCard.image) {
+    imageSrc = defaultSocialCard.image.resize.src
```
kevee, line 36: We want to also set description here to `socialCard.description`

src/components/utils/seo.js
```diff
@@ -2,24 +2,42 @@ import React from 'react'
...
+  } else {
+    imageSrc = socialCard.image.resize.src
+    description = socialCard.description.description
+  }
```
kevee, line 41: Maybe we need to normalize description a bit since Contentful is setting it as `description.description`, but we might pass it as `description` in a manual prop.

schwartzadev: Got it, would you like to restore the `description` prop or use the description value from the `socialCard` prop?

kevee: Let's use the socialCard prop

schwartzadev:
Alright, I think this is good to go.

Here are the Facebook reports (note that the og:url issue is because of the preview site):
image

image

schwartzadev:
...and the Twitter reports:
image
image

kevee:
@schwartzadev it looks like the content page at `/race/about` is missing description:

https://deploy-preview-957--upbeat-lovelace-3e9fff.netlify.app/race/about

src/templates/content.js
```diff
@@ -53,6 +54,16 @@ export const query = graphql`
               }
             }
           }
+          socialCard {
+            description {
+              id
```
kevee, line 59: Should this be `description` instead of `id`?

schwartzadev: Yes! Great eye, thank you. Fixed in 3d6aa0c

PR #965, opened 2020-06-02
Switch to single record queries

### LINKED ISSUE #964: Redo single record queries
**Describe the bug**

Switch from querying the `all` fields to the single ones, when only accessing a single node.

i.e. switch from `allContentfulBlogPost(filter: { id: { eq: $id } }) {` to `contentfulBlogPost(id: {eq: $id})`.

schwartzadev:
## Description
Fixes #964 by switching multi record queries that only return one item to single record queries.


schwartzadev:
@kevee I think this covers all of the cases, but please let me know if there's anything I missed!

schwartzadev:
I messed up the merge conflicts...fixing now.

kevee:
Thanks, @schwartzadev! I also shortened all our queries and iterators by using `nodes` instead of `edges.node`

PR #971, opened 2020-06-03
Focus styles refresh

### LINKED ISSUE #967: Apply new focus style for links across the site:
Let's make focus style consistent across the site [to match this figma file](https://www.figma.com/file/meG3i5VlD94qrTylsqZUXL/Library?node-id=28%3A2)

- [x] Create a links.module.scss file in `/src/scss`
- [x] Add mixins for hover and focus
- [x] Apply mixins to links generally
- [x] Apply mixin to CTA links, press links, all the header and footers for focus

schwartzadev:
## Description
Closes #967, refreshes focus and hover styles for links and subnav/search colors.


tealtan:
This is looking great sitewide! A few additional areas that would be great to iron out:

**The blog bylines on the homepage:**
- Names that are links should have underlines on them, to separate them from names that aren’t links – I think we already do this on the blog posts themselves and the blog index, but not on the homepage.

**Racial Tracker**
- The "show chart data" elements should underline on hover.

Navigation
- The focus style on the logo link should be a white dotted outline, not black.
- Hovering on the top level items should just have the underline –and not the blue bottom border. (I guess the concern here would be how to indicate which top level item a dropdown is for, so I’m thinking that as you mouse around in the dropdown, we style the top level item with a basic underline rather than the blue bottom order.)
- For the dropdown items, I think that the strong background color is enough of an indicator that we make an exception (👻) and not underline them on hover.
- For the Get Involved link, let’s treat it more like a button – so it does not underline on hover (and have no hover style change) and remove the background color change in its focus style.
- On mobile, not sure why but it seems like the top level items and the Get Involved button aren’t getting focus styles?
- On mobile, the hover/focus background color for these subnav items should be white, not black.

Header
- The subnav items here have a blue bottom border on hover – let’s replace with the standard underline style.

**Footer**
- Let’s match the style of these items to the top navigation (bold, no underline until on hover)
- And let’s remove the period at the end of "Back to top"

**Various links around the site**
- I noticed that e.g. on the [About page](https://deploy-preview-971--upbeat-lovelace-3e9fff.netlify.app/about-project) many of the links are now black – which is good, I think that should be the default style – but there are still some links that are plum. Could we go ahead and make all the plum links black? (I think the one exception we may want to make here are the standalone <li> links for each state on the [Our Data page](https://deploy-preview-971--upbeat-lovelace-3e9fff.netlify.app/data), but I’d also be cool using the "CTA" orange link with arrow style instead!)

schwartzadev:
>  I noticed that e.g. on the About page many of the links are now black – which is good, I think that should be the default style – but there are still some links that are plum. Could we go ahead and make all the plum links black? (I think the one exception we may want to make here are the standalone links for each state on the Our Data page, but I’d also be cool using the "CTA" orange link with arrow style instead!)

This makes sense. The current global rules that I set up are as follows:

1. `a` tags are plum
2. `p a` tags are black

I could:
1. Just set all `a` tags to black
2. Expand the (`p a` --> black) configuration to include lists:
```css
p a {
  color: black;
}
li a {
  color: black;
}
```
Which do you prefer @tealtan ?

tealtan:
@schwartzadev Thanks! I think option 1 (setting all a tags to black) is more comprehensive and is what we want – let’s go with that and then we could go through the site and check that it didn’t accidentally clobber something.

schwartzadev:
@tealtan This is ready to go for a second review!

tealtan:
Awesome, I’ll take a look!

tealtan:
> On mobile, the hover/focus background color for these subnav items should be white, not black.

This still looks dark to me!

> The "show chart data" elements should underline on hover.

I’m now seeing these elements always underlined – can we make them only underline on hover, please?

Otherwise, this all looks good, thank you!

schwartzadev:
OK, @tealtan -- we should be good to go. Please do a GH review if this is all set 👍

tealtan:
🎉

PR #984, opened 2020-06-05
Kevee/smart quotes 861

### LINKED ISSUE #861: Add smart quote support to all content pages

kevee:
## Description
Fixes #861 and closes #908

- Adds smart quote component for quoting content
- Wraps contentful content in smartypants
- Updates all our static content to use appropriate HTML entities for smart quotes


kevee:
Good catch, @schwartzadev, fixed those and I think we are good to go.

schwartzadev:
:tada:

PR #1006, opened 2020-06-09
Handle growing list of footer elements

### LINKED ISSUE #996: Handle footer on midsize viewports
With the new T+C link in the footer, we're having some clash at some viewport widths. Here's 814px wide.
image

In this case, let's:
* add right spacing to the logo
* justify the footer links

Once we drop below 768px, like this:
image

...let's also justify the links

schwartzadev:
## Description
Extends @Melvin-Viana's work to fix #996


src/components/layout/footer.module.scss
```diff
@@ -9,16 +9,30 @@
...
   padding: 2rem 0;
 
+  @media (max-width: $viewport-ms) {
+    .footer-list {
+      flex-direction: column;
+    }
+  }
+
   @media (min-width: $viewport-md) {
+    a:first-child {
```
kevee, line 19: Since all the `a` elements are in `li`, they will always be first children. I think we can just remove this, as there was some wrapping in several browsers I tested:

image

schwartzadev: Thanks, good catch! Fixed in 09d2b649d5b17a78324f5aea

Issue #898, opened 2020-05-20
update viz guide


julialedur:
- [x] Delete cdc charts 
- [x] Change top text to text below
- [x] Add new chart and note 

**Delete this text:**

"Complete, up-to-date testing and outcomes data is essential to a successful public health response to the US COVID-19 outbreak, but that data is not being published by the Centers for Disease Control and Prevention (CDC) or any other governmental entity.

The COVID Tracking Project exists to help fill that gap. We collect the best available testing and patient-outcomes data from every US state and territory and the District of Columbia, with full daily updates to the entire dataset." 

**Add this text:** 

"Complete, up-to-date testing and outcomes data is essential to a successful public health response to the US COVID-19 outbreak. For months, we’ve worked to patch together inconsistent state-reported data into a national set of numbers for COVID-19 case, death, and testing in the US.

The CDC has now published a COVID Data Tracker, but their data only partially matches the numbers we get from the state public health authorities as we showed in a [detailed evaluation of the new CDC data.](https://covidtracking.com/documents/CDC_Report_CTP.pdf)."

**Add this chart (Figure 2 [from our recent CDC report](https://covidtracking.com/documents/CDC_Report_CTP.pdf)) after the text.**

Let's add the text as HTML and just the chart as an image, [here's a reference implementation on Observable](https://observablehq.com/d/691c0d6506d4757b).

![](https://p195.p4.n0.cdn.getcloudapp.com/items/kpuLGk87/Screen%20Shot%202020-05-20%20at%2012.01.59%20PM.png?v=275274832b5c2722368f382669c3ca73)

**And this note below the chart:** 

"**Notes:**
Data as of May 15; All units are in absolute numbers and states are only included if the testing count differs by 5% or more.
Our data will always be an undercount. We can only track tests that states report, and not all states report all tests. More significantly, per-capita testing levels in the US remain low, which means that an unknown but probably very large number of people are sick, but aren’t being tested."

kevee:
@julialedur since we aren't using the CDC chart on the homepage or visualization page, can I remove it as a data source for our API?

kevee:
Closed per #1019

PR #1019, opened 2020-06-10
Add the CDC comparison chart to the visualization guide

schwartzadev:
Closes the rest of #898


schwartzadev:
Thanks for the feedback, @julialedur! Let me know how this looks 😄

src/pages/about-data/visualization-guide/index.js
```diff
@@ -280,5 +266,13 @@ export const query = graphql`
         }
       }
     }
+    file(relativePath: { regex: "/cdc-comparison-chart.png/" }) {
+      relativePath
```
goleary, line 270: I think this query is better suited to live within the component as it's currently the only "chart" whose data is fetched within this page (and not in it's own sub component).

goleary:
looks like my review went through after the merge.
None of my comments are blockers, just general suggestions although that typo should be fixed ("discrepencies").

PR #1014, opened 2020-06-09
Standardize typography

schwartzadev:
Adds a consistent type sizing system across the site.

Closes #991

PR #1023, opened 2020-06-12
Add standard sizes for units, spacing, type

kevee:
- Changes all margin, padding, and typography to use standard units
- Includes #1014 for type size
- Auto-imports all mixins from `src/sass`, no need to import mixins
- Re-applies margin and padding spacing across site
- Marks all exceptions with `// ignore-style-rule` for future danger.js integration


.storybook/webpack.config.js
```diff
@@ -33,7 +33,17 @@ module.exports = ({ config }) => {
           localIdentName: '[name]__[local]___[hash:base64:5]',
         },
       },
-      require.resolve('sass-loader'),
+      {
+        loader: require.resolve('sass-loader'),
+
+        options: {
```
kevee, line 39: I'm wondering if we should put this definition in an export or somewhere so we don't have to manage it here and in the webpack config for storybook.

superbiche: That's a great idea. could be exported from gatsby-config.js, or create an external file. I'd rather make it an exported constant in gatsby-config.js for readability.

src/scss/helpers.module.scss
```diff
@@ -0,0 +1,60 @@
+@import './breakpoints.module.scss';
+
+// Base font size. If need be, can be made responsive.
+$base-font-size: 16;
+
+// Spacer keys are the PX values used in the Figma files:
+// 4, 8, 16, 24, 32, 64, 80, 120
+// prettier-ignore
+$spacers: (
```
superbiche, line 9: @kevee should I add a px fallback? To me it seems overkill regarding https://caniuse.com/#search=rem but it's a quick fix to do

tealtan:
**Overall**
- I know we’re not at grid implementation yet, but I’d like to adjust the width on `.layout--content-primary` (at the large breakpoint) from 60% to 66% for now.
- I’d also like to change the default `.layout--content-and-rail .layout--content-primary` font size (at the large breakpoint) to remain at 18px rather than going up to 23px.

**Homepage**
- The See All Data link under latest totals has moved below the stats (it should appear right-aligned at the same level as Latest totals.

**Our Data**
- The 
- The "Totals by state" header should be smaller – it should be 30px.

**Blog Posts**
- Not sure how to describe it, but there’s too much space between the category and the headline.

**Footer**
- On mobile, the footer list shouldn’t have that left-margin.

src/scss/helpers.module.scss
kevee: I think rem is fine.

kevee:
@tealtan just did a few commits that hopefully addressed all these.

src/scss/global.scss
```diff
@@ -189,7 +189,7 @@ Layout blocks
 
 @media (min-width: $viewport-lg) {
   .layout--content-primary {
-    flex-basis: 60%;
+    flex-basis: 606;
```
tealtan, line 192: Should be 66%!

tealtan:
Thanks!

I noticed a typo on the new 66% width, and I’m still seeing too much space below the category and headline:

Screen Shot 2020-06-12 at 6 31 28 PM
Otherwise looks good to me!

src/scss/global.scss
kevee: ha!

kevee:
OK I removed the margin from the top of the blog lede and I think we are good to go. I'm going to do some additional auditing then merge!

PR #1024, opened 2020-06-12
Nav refresh

### LINKED ISSUE #1021: Site header refresh

schwartzadev:
fixes #1021


tealtan:
Two issues with focus-management on the mobile nav:

1. when the nav expands, the search box gets auto-focused. I think the idea behind this (if I’m remembering Ethan’s note right) was that because the nav items appeared before the [Menu] button in the source order, people using screenreaders would not find them unless they moved back. I think we could still achieve this by shifting focus to the nav items div instead?

2. When opening the sub-menus inside the nav, focus moves on the div containing submenu items - which is good – but @Josh Rubenoff and I noticed that "losing focus" (by tapping anywhere on the screen) would close the submenu. This also has the effect of making it so that tapping on the [Menu] button does not collapse the nav – the "losing focus" thing happens, and you have to tap on the [Menu] button a second time to actually close it.

schwartzadev:
Issues to be fixed, from https://github.com/COVID19Tracking/website/pull/1024/commits/e68a17688dff44ef0e5a22b2d84053ebfb8cbd4f

```
- the search *icon* does not open the search bar, only the *input* does
  (this is because the search icon executes the search... the new behavior
  should probably be to have the search icon open the search box when it's
  closed and then execute the search when the box is already open)

- the `display: absolute` used to hide the navigation doesn't render nicely
  when the nav is shown
```

tealtan:
Thanks, this is looking really great.

**A few visual things:**
- [ ] The spacing between nav items could go up from 1.5em to 2ems. (and maybe we want to use rems? It looks the same when I change it, but may be better to be consistent)
- [ ] The color of the dropdown items could be darker at plum-700.
- [ ] The Get Involved buttons should be the same height as the search box/button. (Ideally they are both 40px tall, but as long as they are the same height it’s fine with me.)
- [ ] The padding around the dropdown items can be 0.5em on all 4 sides, rather than 0.5em 1em.
- [ ] On mobile, the Menu/Close text should be vertically centered alongside the logo (it looks bottom-aligned now). Also, the tip of the caret should be pointed towards the middle of the text (it looks like it’s pointing to the left now).

**Search**
- [ ] Clicking on the search icon doesn’t seem to work on Safari or Firefox. (It does work on Chrome.)
- [ ] The search text box should have a border-radius of 4px.
- [ ] The input text style should be font size 1rem, font weight 400, and color #303030.
- [ ] On mobile, the search box label looks too big (should be 13px) and needs more spacing between it and the input text.
- [ ] I’m actually wondering if we should switch to a honey color on the search input focus style. Can we try animating between honey-200 and honey-700 (see [this](https://www.figma.com/file/meG3i5VlD94qrTylsqZUXL/Library?node-id=28%3A3) / [and this](https://codepen.io/tealtan/pen/eYpaPMV) for an example of what I have in mind for the animation)
- [ ] Not sure if this is intentional, but I noticed that when the search input is focused on mobile, the Close link (for the expanded nav) disappears.

schwartzadev:
Thanks @tealtan! I'll tackle this today/tomorrow and reach out if I have any questions!

schwartzadev:
Alright, I think this is ready for you @tealtan!

> Clicking on the search icon doesn’t seem to work on Safari or Firefox. (It does work on Chrome.)

@tealtan can you double-check this one? It seems to be working OK on Browserstack

>  I’m actually wondering if we should switch to a honey color on the search input focus style. Can we try animating between honey-200 and honey-700 (see this / and this for an example of what I have in mind for the animation)

How's this?

tealtan:
Nice, thank you!!!

The search icon seems to work for me now – weird! Maybe I just misclicked last time?

The Get Involved text isn't vertically centered inside the button: setting the line-height to 1.75 seemed to make that better.

Also, the search focus animation should loop back (right now it only goes from dark to light, it should go from light to dark to light and repeat).

schwartzadev:
Alright, @tealtan I think those issues are fixed! I merged in the latest updates, which were a little bit thorny. Please take a look when you can. :100:

schwartzadev:
Ah hello Danger JS...I'll take care of these soon!

tealtan:
Ok great!

I think that the search color animation isn’t as prominent as I was hoping, so let’s remove it and make the box-shadow color `honey-400` instead. Sorry about the back and forth on this one!

Otherwise, this all looks solid to me, thanks so much!

schwartzadev:
OK @tealtan, please approve/merge on GitHub if you're ready!

PR #1031, opened 2020-06-12
Codywall/cleanup state data page links 1026

### LINKED ISSUE #1026: Clean up state data page link language
We had a suggestion to change the state links on pages such as [this](https://covidtracking.com/data) and [this](https://covidtracking.com/data/state/alabama):

Screen Shot 2020-06-12 at 12 18 35 PM

to something more inline and removing repeated language:

Screen Shot 2020-06-12 at 12 22 31 PM
Screen Shot 2020-06-12 at 12 22 37 PM

Maybe someone can answer whether this change should happen on Contentful or if it is in the template? If it is purely within Contentful then we can go with the option above with the vertical separators, if we have template access then it’d be nice to use the separator dots.

codywall:
### Description
Fixes #1026 

- Cleans up state data page link language
- Updated links to be inline and added bullet points between-except on mobile. Font size would need to be decreased on mobile to make all links fit inline, reducing readability.

### DIFF
--- a/src/components/pages/data/state-data.js
+++ b/src/components/pages/data/state-data.js
@@ -23,21 +23,21 @@ const State = ({ state }) => (
     <UnstyledList>
       {state.twitter && (
         <li>
-          <a href={`https://twitter.com/${state.twitter}`}>
-            {state.name} on Twitter
-          </a>
+          <a href={`https://twitter.com/${state.twitter}`}>Official Twitter</a>
+          <span>{'\u2022'}</span>
         </li>
       )}
       {state.covid19Site && (
-        <li>
-          <a href={state.covid19Site}>
-            Best current data source for {state.name}
-          </a>
-        </li>
+        <>
+          <li>
+            <a href={state.covid19Site}>Official Data Source</a>
+            <span>{'\u2022'}</span>
+          </li>
+        </>
       )}
       <li>
         <Link to={`/data/state/${slug(state.name)}#historical`}>
-          Historical data for {state.name}
+          Historical Data
         </Link>
       </li>
     </UnstyledList>
--- a/src/components/pages/data/state-data.js
+++ b/src/components/pages/data/state-data.js
@@ -3,7 +3,7 @@ import marked from 'marked'
 import smartypants from 'smartypants'
 import { Link } from 'gatsby'
 import slug from '~utilities/slug'
-import { UnstyledList } from '~components/common/lists'
+import StateLinks from '~components/pages/state/state-links'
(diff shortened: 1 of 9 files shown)


src/components/pages/data/state-data.js
```diff
@@ -23,21 +23,21 @@ const State = ({ state }) => (
     <UnstyledList>
       {state.twitter && (
         <li>
-          <a href={`https://twitter.com/${state.twitter}`}>
-            {state.name} on Twitter
-          </a>
+          <a href={`https://twitter.com/${state.twitter}`}>Official Twitter</a>
```
kevee, line 26: To help with screen reader users and prevent duplicate links, let's rewrite these as:

```
<span className="a11y-only">{state.name}'s</span>Official Twitter
```

src/components/pages/data/state-data.js
```diff
@@ -23,21 +23,21 @@ const State = ({ state }) => (
...
       )}
       {state.covid19Site && (
-        <li>
-          <a href={state.covid19Site}>
-            Best current data source for {state.name}
-          </a>
-        </li>
+        <>
+          <li>
+            <a href={state.covid19Site}>Official Data Source</a>
```
kevee, line 33: Ditto on using a11y-only to keep the state name

src/components/pages/data/state-data.js
codywall: Good idea, I'll add that.

codywall:
Sounds good @kevee, I added the accessibility spans and will refactor the `StateLinks` component to work in `state-data.js`.

src/components/pages/data/state-data.js
```diff
@@ -24,20 +24,26 @@ const State = ({ state }) => (
...
-        <li>
-          <a href={state.covid19Site}>
-            Best current data source for {state.name}
-          </a>
-        </li>
+        <>
+          <li>
+            <a href={state.covid19Site}>
+              <span className="a11y-only">{state.name}&apos;s</span>
+              Official Data Source
```
kevee, line 38: Let's change this to "Best Current Data Source"

kevee: Also, thanks so much for this, the pages are looking way shorter!

codywall: No problem! Thanks for the feedback.

kevee:
Looking good! I just:

- Changed the style to be it's own module
- Changed dataSource (not used anywhere) to our secondary site
- Updated style to use our new type/spacer system
- Fixed error in test file from waay back when.

Issue #1033, opened 2020-06-12
Surface downloadable CSVs


kevee:
In order to make our CSVs more accessible, let's:

- [x] Update the preamble of the data page:
> We update all the data each day between 4pm and 5pm EDT. You can also get the dataset as a [spreadsheet](), [CSV files](), and an [API]().
- [x] Add a link to `/data/download` in the preamble of the `/api` page

## Create a new page at `/data/download` that:

- [x] Shows similar information about caveats that we use in the API page.
- [x] Lists US and state download links
- [x] Displays all the fields, along with descriptions, like we do with the API page

### Page structure:

- US history
  - Download all history data
  - Field descriptions (collapsible)
- State history
  - Download State data
    - Field descriptions (collapsible)
    - Download all states (`/api/v1/states/daily.csv`)
    - Download single state (collapsible)
      - List all state files (`/api/v1/states/[state]/daily.csv`)

## Discussion:

I think the [API explorer](https://covidtracking.com/api) could still work well for downloadable CSV files, but I could be wrong. FYI all the fields and descriptions are defined by the API and automagically available, so reformatting the display is easy.

We *could* provide a list of all the downloadable CSV files by state (the daily history). I don't think linking off to the `api/v1/states/[state]/[date].csv` files makes sense.

amandafrench:
I think the word “spreadsheet” should be replaced with “read-only spreadsheet” or similar maybe — and maybe make the downloadable CSV files the first thing in the sentence.

kissane:
Maaaaaybe

> We update all the data each day between 4pm and 5pm EDT. You can also [see the dataset as a spreadsheet](), [download CSV files for everything we track](), or use an [API]().

kissane:
> We update all the data each day between 4pm and 5pm EDT. You can also [download CSV files for everything we track](), [see the dataset as a spreadsheet](), or use our [APIs]().

is "APIs" correct

amandafrench:
I myself would rather like a list of per-state downloadable CSVs on a new Download page since I’m mainly interested in N.C.

amandafrench:
And am not really an API user

kevee:
I think the API page is fine, but we should make the download page it's own thing. I updated the page structure accordingly.

amandafrench:
Love the preview I saw! One question: are the national and state-level field descriptions going to be automatically updatable? Because if not I almost think that putting links there to the API or wherever they are primarily maintained might be better. 

[...]

PR #1040, opened 2020-06-16
feature(downloads): adds historical us and state csv downloads

scandycuz:
Adds feature for issue #1033 (Surface downloadable CSVs)

Please let me know if anything should be done differently. One thing I still had some questions about is updating the GraphQL content, such as the preamble on the data page. I imagine it's in a database somewhere? 

I was also unsure how to add the '← Our Data' navigation button above the page title, although I was thinking it relied on some GraphQL data as well.


kevee:
Thanks, @scandycuz! Here's my thoughts:

The preamble is actually stored in Netlify, so if you add this component:

```
<ContentfulContent
      content={
        data.contentfulSnippet.childContentfulSnippetContentTextNode
          .childMarkdownRemark.html
      }
      id={data.contentfulSnippet.contentful_id}
    />```

Then this query to the page:

```
contentfulSnippet(slug: { eq: "download-page-preamble" }) {
      contentful_id
      childContentfulSnippetContentTextNode {
        childMarkdownRemark {
          html
        }
      }
    }
```

You should be good to go!

For the "return link," set the `navigation` prop on the `Layout` to: `{data.contentfulNavigationGroup.pages}`

Then add this query to the page to snag the "data" navigation group:

```
contentfulNavigationGroup(slug: { eq: "data" }) {
      pages {
        ... on ContentfulPage {
          title
          link: slug
        }
        ... on ContentfulNavigationLink {
          title
          link: url
        }
      }
    }
```

**Comment by scandycuz** [comment]
Thanks @kevee! I've made the changes. 

It does seem that the 'return link' still doesn't display, but I think it may just need to be added to the GraphQL data?

**Comment by amandafrench** [comment]
Thanks so much for doing this much-needed work, @scandycuz!

**Comment by scandycuz** [comment]
@amandafrench of course, no problem!

**Comment by kevee** [comment]
Suggestions from @amandafrench:

- [ ] "US History" --> "US Data" (or possibly "US COVID-19 Data")
- [ ] "Download all history data" --> "Download all available US data"
- [ ] "State History" --> "State Data" (or possibly "State COVID-19 Data")
- [ ] "Download all states" --> "Download all available data for all states"
- [ ] "Download single state" --> "Download all available data for a single state"

**Comment by scandycuz** [comment]
@kevee Happy to make the changes. Also, currently the headers are hard coded, but should I link them to Contentful instead? If so, I can add the content there, or just link to it if it would be easier for someone else to create it.

**Comment by kevee** [comment]
Let's make them hard-coded, the contentful part should just be the top part of the page.

**Comment by scandycuz** [comment]
Sounds good, thanks

File: src/pages/data/download.js [hunk 1]
```diff
@@ -0,0 +1,82 @@
...
+      id={data.contentfulSnippet.contentful_id}
+    />
+
+    <h2>US Data</h2>
+    <DownloadLink
+      desc="Download all available US data"
+      path="/api/v1/us/daily.csv"
+    />
+    <Panel label="Field Descriptions">
+      <Fields schema="Us" />
```
scandycuz, line 29: @kevee Is it possible that the name of the schema could change? If so I could get the schema name from the path, e.g.
```
import definition from '../../../_api/v1/openapi.json'
...
  <Fields
    schema={
      definition.paths['/api/v1/states/daily.{format}'].get.responses[200].content['text/csv']
        .schema.items.$ref.split('/').pop()
    }
  />
...
```

PR #1044, opened 2020-06-16
chore(docs): Change fail message in plugin

### LINKED ISSUE #1042: missing api data error message is no longer useful
**Describe the bug**
Perhaps I'm just experiencing this because I had a very old version of the repo on my machine, but 
`setup:api-repo` was "succeeding" but not actually downloading the needed files.

I got the below error message which if I remember correctly describes quite an old way of getting the api data onto your machine.

```
 ERROR 

There is no file ./_api/v1/internal/race-homepage.json directory.

If you are developing on your local computer, be sure to download the latest data archive here: http://covidtracking.com/api/archive.tar.gz. Then expand it to the directory "_data".
```

I was able to fix it by running `git clone --depth 1 https://github.com/COVID19Tracking/covid-public-api.git _api && rm -rf _api/.git` without the preceeding commands in `setup:api-repo`.

Environment: Windows 10

kevee:
- Changes the error message that is displayed if a user tries to run gatsby without the API files downloaded
- Fixes #1042

### DIFF
--- a/plugins/gatsby-source-covid-tracking-api/gatsby-node.js
+++ b/plugins/gatsby-source-covid-tracking-api/gatsby-node.js
@@ -12,9 +12,9 @@ exports.sourceNodes = async (
     fs.statSync(configOptions.file)
   } catch {
     reporter.error(
-      `There is no file ${configOptions.file} directory. 
+      `There is no file ${configOptions.file}. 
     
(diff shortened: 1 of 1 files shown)


goleary:
well that was fast...makes me feel like I should've just fixed it myself...thanks!

Issue #987, opened 2020-06-05
Design system: Forms


kevee:
[Figma file](https://www.figma.com/file/meG3i5VlD94qrTylsqZUXL/Library?node-id=28%3A3)

Let's create common components for all forms that aims for consistency and accessibility. There are already some components in `/src/components/common/forms`.

- [x] Create a single component for field groups (accepts a `label`, `isRequired`, and `field` component prop). Required fields get a readable 'required' label

This should then render something like:

```
<div class...>
<label for={id}>{label} <span>Required</span></label>
{children}
</div>
```

- [x] Create components for common input elements that use that field group. Something like:

```
const FormInput = ({type, name, label, isRequired, onChange}) => (
  <FormGroup label={label} isRequired={isRequired} id={name}>
  <input type={text} name={name} id={id} onChange={onChange}/>
</FormGroup>
)
```

Create one of these for:
- [x] Regular input field, with configurable `type` prop, i.e. `<FormInput type="text">` would render an `<input type="text">` in a field group.
- [x] Textarea
- [x] Radio & checkbox (maybe we just pass an array of values?)

- [x] We also need to pass events to these as well, let's do `onChange` for now. There is only one form that monitors inputs.

- [x] Support for half-width form groups

### Update forms:

- [x] Volunteer form
- [x] Contact form
- [x] Get involved form
- [x] Search page

tiagofsanchez:
Think this could be something like this for the `input` component: 

```
const input = props => {
  const { inputtype, label } = props;

  let inputElement = null;
  switch (inputtype) {
    case "input":
      inputElement = <input className={inputStyle} {...props} />;
      break;
    case "text":
      inputElement = <textarea className={inputStyle} {...props} />;
//you can have as may cases you want here
      break;
    default:
      inputElement = <input sx={inputStyle} {...props} />;
  }

  return (
    <div className={theDivStyle}>
      <p className={theLabelStyle}>{label}</p>
      {inputElement}
    </div>
  );
};

```

Like this the Input component will be just <Input inputElement={input} onChange={changeHandler} />

Thoughts?

tiagofsanchez:
Started to work on this and created a new branch [here](https://github.com/COVID19Tracking/website/tree/form-design-system). 

Would be great to have someone to review to make sure that it make sense and improves the current implementation. 

Changes made: 
- The FormInput now as a switch statement where we can defined the `inputElement` that we want, we could define anything here
[...]

PR #1047, opened 2020-06-17
Form design system, Design system: Forms issue #987

tiagofsanchez:
Design system: Forms #987

Implementing the new Form design system


kevee:
Hi, @tiagofsanchez! I merged in master and updated the SCSS a bit to use our new spacers and type size mixins.

tiagofsanchez:
Will check this out @kevee and will update the rest

tiagofsanchez:
@kevee, I reckon you have adjusted almost everything. I did minor things. Whenever you have time you can review. After I can document on storybook

kevee:
Looks good! Let's get that storybook implemented then I think we are GTG!

kevee:
Thanks, @tiagofsanchez! Only issue I'm seeing is the react recaptcha component is throwing an error when you:

1. Click on contact
2. Click on volunteer form, or really any page
3. Hit the back button

I think it's probably [due to this issue](https://github.com/appleboy/react-recaptcha/issues/181), but it's weird that this isn't already happening on the live site. I'll have time later today to try and debug it as well.

tiagofsanchez:
@kevee, yes this is strange. I have done a couple of debugging but still don't get the problem: 
- in dev mode works well
- in chrome on my phone i have no issues, only on desktop chrome
- without the recaptcha it works well, so that is def the issue
- implemented the `render='explicit'` and `onloadCallback` props and not fixed yet
Will try latter tonight again

kevee:
I broke some of these form elements into separate components and reverted search since it's a distinct form element that can be inconsistent with the others.

PR #1054, opened 2020-06-18
feat(state-charts): update bar chart component & add prelim charts to state data pages.

goleary:
- made some small improvements to the BarChat component
- added bar charts for test, positive & death increase metrics to each state page.

some things to be done:
- [ ] moving avg line
- [ ] labels
- [ ] alignment

### DIFF
--- a/src/components/pages/state/charts/index.js
+++ b/src/components/pages/state/charts/index.js
@@ -1,5 +1,20 @@
 import React from 'react'
 
+import BarChart from '../../../charts/bar-chart'
+
+import {
+  deathsBarColor,
+  positiveColor,
+  totalColor,
+  parseDate,
+} from '~utilities/visualization'
+
+import chartsStyles from './charts.module.scss'
+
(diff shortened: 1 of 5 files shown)


kevee:
Thanks, @goleary! I made one change to prevent sorting from affecting the history table. I'm going to merge this into the feature branch so we have a consistent place for this work to be merged into.

src/components/pages/state/charts/index.js
```diff
@@ -14,11 +26,51 @@ const dailyAverage = (history, field, range = 7) => {
...
+  return data.map(d => ({
+    date: parseDate(d.date),
+    value: d[field],
+  }))
+}
+
+// we need a combined bar + line chart
+
+export default ({ history }) => {
+  const data = [...history].sort((a, b) => a.date - b.date)
```
kevee, line 44: This sort is being applied to the parent history array. Let's clone it so it won't also sort the history table.

Issue #1058, opened 2020-06-19
Clicking away from window should close nav search bar


scandycuz:
## Replicate
Checkout the `feature/nav-refresh` branch (see #1024 )
1. Open the search bar
2. Click outside the document

## Expected behavior
Search bar should close

## Actual Behavior
Nothing happens

Issue #1057, opened 2020-06-19
Nav search input closes when 'SEARCH' label clicked


scandycuz:
## Replicate
Checkout the `feature/nav-refresh` branch (see #1024 )
1. Open the search bar
2. Click the 'SEARCH' label

## Expected behavior
Nothing happens

## Actual Behavior
The search bar closes, and in Firefox, the Navigation items don't reappear

PR #1059, opened 2020-06-19
Bugfixes: nav bar 'SEARCH' click and away from window click

scandycuz:
- Fixes issue #1057 (Nav search input closes when 'SEARCH' label clicked)
- Fixes issue #1058 (Clicking away from window should close nav search bar)

### DIFF
--- a/src/__tests__/components/layout/__snapshots__/header.js.snap
+++ b/src/__tests__/components/layout/__snapshots__/header.js.snap
@@ -27,7 +27,8 @@ exports[`Components : Layout : Header renders correctly 1`] = `
         }
       >
         <div
-          className="searchInput"
+          className="searchInput "
+          onBlur={[Function]}
         >
           <div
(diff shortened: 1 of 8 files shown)


scandycuz:
@schwartzadev It seems that the search label bug was caused by the CSS `:focus` styling and the styling based on the `autocompleteHasFocus` variable getting out of sync due to Firefox handling `:focus` differently.

I've updated the styling to always rely on the `autocompleteHasFocus` instead of `:focus`, and it seems to work better now.

I also added an `onBlur` handler for when the user clicks away while the search bar is open

scandycuz:
@schwartzadev I think the 'SEARCH' bug was occuring due to some conflicts between styling that was applied from the `:focus`selector and styling that was applied when `autocompleteHasFocus` was `true`. I updated the styling to always reference `autocompleteHasFocus`, and it seems to work well now.

PR #1063, opened 2020-06-19
Use grid styles for the footer

### LINKED ISSUE #1009: Update the footer design now that it’s got more links in it
Screen Shot 2020-06-08 at 9 31 27 PM

Has a few more links and an attribution added recently, we should update the design here to fit things better.

schwartzadev:
Implements @tealtan's grid (based off of @kevee's work on #1056) and @jhuashao's designs.
* Closes #1009
* Uses the new grid styles in the footer

Note that I cheated a little bit with *css* grid for the footer links, but I hope that's OK.

@jhuashao: please review when you get a chance and share any changes to be made in this PR.

## ~~Pre-merge issues:~~
The mocks have an RSS link in the footer, but we don't have that link in the footer now. At the moment, I just left a todo, but we should either:
1. Link this to something else
2. Remove this link from the footer
3. Add the correct link to RSS if we support it


src/components/layout/footer.module.scss
```diff
@@ -1,99 +1,143 @@
...
-      &:hover {
-        @include link-hover-dark-bg;
-      }
+  .social {
+    @include col(1 3 2);
+    @include margin(0, (top, bottom));
+
+    display: flex;
+    justify-content: right;
+
```
kevee, line 89: Do we want the social icons to float right on the medium breakpoint?
image

src/components/layout/footer.module.scss
```diff
@@ -1,99 +1,143 @@
...
-      @include link-focus-dark-bg;
+  }
+
+  .back-to-top {
+    @include col(1 1 2);
+
+    display: flex;
+    align-items: center;
+    justify-content: right;
+
```
kevee, line 131: Let's also have the back-to-top link float to the right below lg breakpoints so it's flush.

kevee:
[Here's the active branch preview](https://schwartzadev-footer-redux--upbeat-lovelace-3e9fff.netlify.app/)

kevee:
On the question of the RSS feed, which I totally missed! We currently don't support RSS feeds, but could [easily do so with this plugin](https://www.gatsbyjs.org/packages/gatsby-plugin-feed/).

However, I feel like that is a separate discussion and issue, and I think out of scope for this PR. Let's remove the RSS link for now from this PR, and the empty, blank, lonely space will forever yell at me to create a new issue and get this going.

src/components/layout/footer.module.scss
schwartzadev: I believe so, and I think they `float: right;` now. Do you want to change that behavior?

src/components/layout/footer.module.scss
schwartzadev: I think this is already the case. Here's a screenshot:

image

src/components/layout/footer.module.scss
kevee: Hm, this is what I'm getting on Chrome:

image

tealtan:
Oh, thanks, I’d missed that! (Re: RSS). Yes, agree that it’s for another time.

src/components/layout/footer.module.scss
schwartzadev: Fixed in 23cf28d0b3dba16f8f52f0994cdb412d44c1f7f0 -- see the commit message for details

src/components/layout/footer.module.scss
schwartzadev: Fixed in 23cf28d0b3dba16f8f52f0994cdb412d44c1f7f0 -- see the commit message for details

jhuashao:
This looks great to me! All breakpoints seem to be working well, and I don't see any visual issues at the moment.

Ready to ship! 🚀

PR #1076, opened 2020-06-22
fix(summary-table): last updated times

### LINKED ISSUE #1075: "Last updated (ET)" times are -4 hours wrong on data pages
On https://covidtracking.com/data and on state pages such as https://covidtracking.com/data/state/delaware there are "Last updated" times underneath each state's main data table, and all of them seem to be 4 hours behind what they should be. Delaware's time zone is Eastern, for example, and according to [our spreadsheet](https://docs.google.com/spreadsheets/u/2/d/e/2PACX-1vRwAqp96T9sYYq2-i7Tj0pvTf6XVHjDSMIKBdZHXiCGGdNC0ypEU9NbngS8mxea55JuCFuua1MUeOj5/pubhtml#) its last update time in Eastern was 6/20 18:00 (see both States Current and States Daily). But the data pages report its last update time as 6/20 at 14:00 (i.e., 2pm). Same is true for all other states I've spot-checked. 
Screenshot 2020-06-22 08 31 27

I also think it would be clearer if that timestamp were labeled something like "State's reported update time" and the one in the blue box were labeled something like "This page last synced with our spreadsheet."

schwartzadev:
This fixes the time offset on the summary tables, identified in #1075.

The issue is that the API returns datetimes that are _already in Eastern Time_, yet have a UTC timezone.

This solution is admittedly hack-y...we should probably fix this upstream (i.e. in the API build) so that our API customers also don't have this problem

Closes #1075.

### DIFF
--- a/src/__tests__/components/common/__snapshots__/summary-table.js.snap
+++ b/src/__tests__/components/common/__snapshots__/summary-table.js.snap
@@ -257,7 +257,7 @@ exports[`Components : Common: Publication: Summary table renders correctly 1`] =
     className="detail-text detailText"
   >
     Last updated: 
-    Wed Apr 1 2020 12:00 am
+    Wed Apr 1 2020 4:00 am
      
     <abbr
       aria-label="Eastern Time"
@@ -406,7 +406,7 @@ exports[`Components : Common: Publication: Summary table renders correctly 2`] =
     className="detail-text detailText"
   >
     Last updated: 
-    Wed Apr 1 2020 12:00 am
+    Wed Apr 1 2020 4:00 am
      
     <abbr
       aria-label="Eastern Time"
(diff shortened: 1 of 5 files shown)


schwartzadev:
@amandafrench please take a look at the preview site and confirm that this fixes #1075. Thank you! :tada:

kevee:
I'm going to see if we can't just fix this in the API now. Hold on this PR for now.

kevee:
@schwartzadev the issue is we are using a deprecated `dateModified` field from the API. Let's switch to `lastUpdateEt`, which is an appropriately-labeled datetime.

amandafrench:
The Last Updated times in Eastern on the web preview site do now match the Last Updated times in Eastern from the spreadsheet, so that works. We can hold off on the relabeling if need be.

schwartzadev:
@kevee it looks like that field returns dates like `6/19/2020 18:00`. Can we update this in the API to match our style guide with the lowercase meridiem?

Meantime, I'm working on a new PR to use the `lastUpdateEt` field.

schwartzadev:
Closing this in favor of #1104, which uses the `lastUpdateEt` field

PR #1079, opened 2020-06-22
feat(state-data): Add optional line to Bar Graph (avg)

goleary:
- added Charts to main data page (because I like it) - we will want to
reorganize the component if we decide to stick with this
- added 7 day rolling average to the bar charts
- fixed a couple bugs in dailyAvg function
- added hospitalization data where available.
- added titles

### DIFF
--- a/src/components/pages/state/charts/index.js
+++ b/src/components/pages/state/charts/index.js
@@ -44,14 +44,12 @@ const getDataForField = (data, field) => {
 // we need a combined bar + line chart
 
 export default ({ history }) => {
-  // change below
+  // Change below to update range of chart
   const NUM_DAYS = 90
   const data = [...history]
     .slice(0, NUM_DAYS)
     // .slice(history.length - NUM_DAYS, history.length)
     .sort((a, b) => a.date - b.date)
-  // eslint-disable-next-line no-debugger
-  debugger
 
   const hasHospitalizationData = history[0].hospitalized !== null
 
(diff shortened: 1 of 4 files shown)


src/components/pages/state/charts/index.js
```diff
@@ -41,7 +44,15 @@ const getDataForField = (data, field) => {
 // we need a combined bar + line chart
 
 export default ({ history }) => {
-  const data = [...history].sort((a, b) => a.date - b.date)
+  // Change below to update range of chart
+  const NUM_DAYS = 90
```
kevee, line 48: In the future, let's move these kind of variables to `gatsby-config.js`'s site metadata

src/components/pages/state/charts/index.js
```diff
@@ -55,22 +66,44 @@ export default ({ history }) => {
...
-        data={getDataForField(data, 'positiveIncrease')}
-        fill={positiveColor}
-        {...props}
-      />
-      <BarChart
-        data={getDataForField(data, 'deathIncrease')}
-        fill={deathsBarColor}
-        {...props}
-      />
+      <div>
```
kevee, line 69: After I merge this into the feature branch, I'll update that branch from master to give you access to our new [grid system](https://covid19tracking.github.io/website-docs/?path=/story/layout--grids) to make these charts responsive. I would say we want these as four charts (25% wide) on large breakpoints, two (50%) on medium, and one (full width) on small.

src/components/pages/state/charts/index.js
goleary: sounds good, will do.

src/components/pages/state/charts/index.js
goleary: lovely, that will be much preferred to the size hard coding that's been done so far.

PR #1104, opened 2020-06-26
Fix summary table times

### LINKED ISSUE #1075: "Last updated (ET)" times are -4 hours wrong on data pages

schwartzadev:
closes #1075

### DIFF
--- a/src/__tests__/components/common/__snapshots__/summary-table.js.snap
+++ b/src/__tests__/components/common/__snapshots__/summary-table.js.snap
@@ -256,8 +256,8 @@ exports[`Components : Common: Publication: Summary table renders correctly 1`] =
   <div
     className="detail-text detailText"
   >
-    Last updated: 
-    Wed Apr 1 2020 12:00 am
+    State’s last reported update time: 
+    20200401
      
     <abbr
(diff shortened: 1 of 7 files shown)


schwartzadev:
@amandafrench Does this fix work for you?

PR #1107, opened 2020-06-26
chore(nav): center get involved button on small viewports

schwartzadev:
The text inside the "Get Involved" button is not vertically centered at smaller viewports.

Currently, it looks like this:
image

This PR centers the text:
image

### DIFF
--- a/src/components/layout/header/header.module.scss
+++ b/src/components/layout/header/header.module.scss
@@ -24,7 +24,9 @@
 
 .get-involved {
   display: inline-block;
-  line-height: 1.75;
+  @media (min-width: $viewport-lg) {
+    line-height: 1.75;
+  }
   padding: 0.4rem 0.5rem; // ignore-style-rule
   transition: all 0.2s;
   height: 35px; // match the search box height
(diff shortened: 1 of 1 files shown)


src/components/layout/header/header.module.scss
```diff
@@ -24,7 +24,9 @@
 
 .get-involved {
   display: inline-block;
-  line-height: 1.75;
+  @media (min-width: $viewport-lg) {
+    line-height: 1.75;
+  }
   padding: 0.4rem 0.5rem; // ignore-style-rule
   transition: all 0.2s;
   height: 35px; // match the search box height
```
tealtan, line 32: @schwartzadev I was thinking we could move this explicit height to the viewport-lg and above media query above, since we only need to match the search box height at that range? I think this will help the vertical centering at the smaller breakpoints as well.

tealtan:
Thanks! See [my comment](https://github.com/COVID19Tracking/website/pull/1107/files#r446321880) but otherwise this is fine with me!

PR #1108, opened 2020-06-26
chore(blog): wrap author byline on small devices

schwartzadev:
:new: 
image

These bylines currently don't wrap the date to a new line, like this:
image

@tealtan the date looks a little lonely on it's own. Do you have any suggestions for that? Maybe we could darken the color or prefix it with "Published on" :thinking:...

### DIFF
--- a/src/__tests__/components/pages/blog/__snapshots__/blog-lede.js.snap
+++ b/src/__tests__/components/pages/blog/__snapshots__/blog-lede.js.snap
@@ -15,30 +15,22 @@ exports[`Components : Pages : Blog : Lede renders correctly 1`] = `
   <div
     className="byline"
   >
-    <p
-      className="bylineText"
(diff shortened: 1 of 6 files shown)


tealtan:
Oh cool! – I was actually fiddling with this last night - what do you think of this?
https://www.figma.com/file/GwkYSPVlTkdgLyGNgeMrLz/v3-Refresh?node-id=548%3A7014

- We wrap the byline and date and adjust the colors (use #000 for byline and slate-600 for the date)
- Above mobile: byline photos don’t overlap each other, but are spaced 8px apart.
- We make the byline and date font size 13px.
- We make byline links regular weight (not bold), as the underline is enough.

schwartzadev:
I'm going to move this to the redesign per #1109. Thanks @tealtan!

PR #1119, opened 2020-06-26
Blog post redesign, center content, and about-page redesign

### LINKED ISSUE #1109: Blog post redesign
[Figma file](https://www.figma.com/file/GwkYSPVlTkdgLyGNgeMrLz/v3-Refresh?node-id=105%3A1278)
Note this issue depends on #1110 

Implement a new design for blog pages as outlined in the figma:

- [x] Create a new prop for `<Layout>` that accepts a hero element
- [x] Move the title, lede, authors into that component

## Add a new component for social sharing
Let's make them generic so we can re-use later, and make sure each one is it's own component:
- [x] Twitter - `https://twitter.com/intent/tweet?url=[url]&text=[lede]`
- [x] Facebook - `https://www.facebook.com/sharer/sharer.php?u=[url]`
- [x] Link - Going to need to use state to say "did user click button." When user clicks, use [copy-to-clipboard](https://www.npmjs.com/package/copy-to-clipboard) to copy the page URL, and maybe change the icon to a thumbs up? Show a success message? We'll get back to you on this.

schwartzadev:
Closes #1109 

Merge in changes to centering content from #1130 
Merge in changes to about page #1121 

### Pre-merge
- [x] Choose and implement behavior for copying a link
- [x] Mock feature image styles, like `blog/its-not-just-testing`


tealtan:
This is looking great, @schwartzadev!

A few notes:

- [ ] The width of the blog header should match the width of the content. Their width at 1200px and above looks right to me, but below there seems to be some variation. I noticed that there are two `.container-module--container` classes being used in the header, which is effectively doubling the amount of margin that should be there. (When I added some inline styles to set the inner container’s margins to 0, things looked lined up to me – you’ll know better than me what the right solution is, this is just me trying to properly explain what I’m seeing since it’s a bit weird!)

- [ ] I wonder if we ought to show the 3 recent posts regardless of category (rather than by the current category) since we currently don’t have many blog posts in many categories. Or, if this isn’t too complicated, we could show 3 recent posts by category if there are at least 3, otherwise we fall back to regardless of category?

- [ ] Bylines should be 13px for their font size.

- I’ve noticed a few times on the site where byline images seem to be stretched – could we explicitly set them to 40px wide and 40px tall? (Ok with me to set them to 2.5rem to allow them to scale). And as mentioned in Slack, we can actually show them on mobile – my blueberry note was for a previous iteration of the design, sorry! (If we are able to support copy-to-clipboard on mobile, feel free to add that back as well!)

Thanks!

schwartzadev:
Perfect, thanks for the feedback @tealtan! I'll tackle this tomorrow morning/mid-day.

kevee:
Note that these three PRs are going to be merged all together: #1130, #1119 , #1121

samWinfield:
@schwartzadev Looks great! I have a few small style suggestions that I'll post below:

 - [ ] I think we should add a full-width top rule to the related posts section, using the style that is used on the `homepage: 1px solid #d2d6d7` (The placement and spacing around the rule should be [in the figma file here](https://www.figma.com/file/GwkYSPVlTkdgLyGNgeMrLz/v3-Refresh?node-id=628%3A7314). I think it would be good to have some separation between the main content and the related posts.)
 - [ ] It looks like the margins between the related post modules are not collapsing. The spacing there should match the spacing between the title ("Related Posts") and the first related post. (It's currently rendering at 64px/4rem on desktop, and it should be 32px/2rem).

These next two things are nit-picky, and not urgent if they're more than a quick fix.

 - [ ] For the byline module, if it doesn't effect other styles, I would change the line-height to 1.2, to tighten up the spacing there.
 - [ ] And for the last paragraph in the blog post, change margin bottom to 3rem  (It's currently 1.5).

Thank you!

kevee:
On the link sharing button, I think when you click on it, just copy to the clipboard with [copy-to-clipboard](https://www.npmjs.com/package/copy-to-clipboard) and then show a small success feedback (use an aria polite alert)

image

tealtan:
Thanks! And good reminder for me to give that success feedback a touch of design work.

schwartzadev:
OK @tealtan and @samWinfield Kevin and I combined the author page updates and the blog page updates into this PR. I hope that's not a big issue for you! Please take a second pass 🔍 at the preview site when you have a chance!

tealtan:
Yay! A few more notes:

- [ ] There’s double margin coming from the `li` and the `p` elements inside - [see screenshot](https://covid-tracking.slack.com/files/UUW27UX2M/F016KFY44S0/screen_shot_2020-07-01_at_12.54.14_pm.png)
- [ ] The author name and date should be 13px.
- [ ] Can we swap out the ampersand with an "and"?
- [ ] The horizontal rule in the footer should go full-width - [see screenshot](https://user-images.githubusercontent.com/293844/86271375-c4d6f180-bb9a-11ea-9dc2-d162537b8eed.png)
- [ ] I really think we can remove the underlines on the headlines in the Related posts area.
- [ ] Can we have equal spacing above and below the horizontal rule on related posts? It should be 4rems above 768px and 3rems below 768px. (We are standardizing on this kind of vertical spacing to divide sections of a page, so it may be worth creating a reusable thing in our CSS.)
- [ ] There should be spacing below the image (we can match the spacing that’s on paragraphs) - [see screenshot](https://user-images.githubusercontent.com/293844/86271775-7fff8a80-bb9b-11ea-9017-b3340843b9c7.png)

Lastly, can we add some styling to the copy link message that appears after clicking on the button? Something like this:
https://www.figma.com/file/GwkYSPVlTkdgLyGNgeMrLz/v3-Refresh?node-id=779%3A7585

This is looking so close, thank you!

PR #1121, opened 2020-06-26
About page redesign

### LINKED ISSUE #1103: About page redesign
[Figma file](https://www.figma.com/file/GwkYSPVlTkdgLyGNgeMrLz/v3-Refresh?node-id=472%3A6459)

Change `/src/pages/about-project/index.js` to:

- [x] Remove `narrow` from layout so the contributor container is wider
- [x] Wrap the contentful content up top and bottom with containers with `narrow` props
- [x] Change the contributors section to 6 cols on large breakpoints, 2 on medium, and 1 on small.

schwartzadev:
Closes #1103


tealtan:
Thanks Andrew! A few additional tweaks:

- [ ] Can we make the team leads list two columns above the 768px breakpoint?
- [ ] Can we add 1rem of padding to the right of the volunteer list’s <li> items so that they have a bit more space between each column?
- [ ] Can we add 1rem of space between the <li> items? This can apply to all 3 lists on this page (maybe a global style)?
- [ ] Ideally, the `Are you a volunteer who wants to be added to this list? Edit your profile in Slack and scroll down to "Volunteer web page."` appears as a standard paragraph that is less wide than the volunteer list, but this is not a big deal either way.

**Not sure if this is for later**
I noticed other items that maybe we want to address separately in #1129, if this PR is just for the About page?
- [ ] I noticed that the `h2`s are 23px below 1080px and 28px above 1080px – could we adjust this to be 30px below 768px and 36px above 768px? Also, they should have 2rems of bottom margin (on all viewport sizes).
- [ ] The header bottom margin is currently always 3rem – I’d want this to continue to be 3rem below 768px, and 4rem above 768px. (And apply this same spacing as top margin for the footer?)
- [ ] The header should be 8/12 columns and centered, matching the body content.

kevee:
Note that these three PRs are going to be merged all together: #1130, #1103, #1109

schwartzadev:
Thanks so much for the feedback! Just a couple things @tealtan

>  Can we make the team leads list two columns above the 768px breakpoint?

This will be tricky -- this comes from markdown in Contentful...are you OK with leaving it as one column?

>  The header bottom margin is currently always 3rem – I’d want this to continue to be 3rem below 768px, and 4rem above 768px. (And apply this same spacing as top margin for the footer?)

I'm having trouble finding this margin -- can you take a screenshot and show me where this margin is?

kevee:
Let's leave the team leads alone for now, I like that it's easily editable in Contentful.

schwartzadev:
@tealtan I found the header margins and fixed them up in the redesign feature branch.

tealtan:
Re: team leads being in Contentful – ah, yep, that makes sense!

tealtan:
And awesome – that margin change looks right to me in that branch. Thank you!

schwartzadev:
I did this merge on my development machine but GitHub didn't seem to register it.

PR #1130, opened 2020-06-28
chore: Change all narrow containers to centered

### LINKED ISSUE #1129: Change all narrow page layouts to 'centered'
- [ ] Change all the pages using `narrow` layouts to `centered`
- [ ] See if it's possible to merge the homepage/CRDT homepage containers with the new `narrow` Container prop

kevee:
- Fixes #1129 
- Does not alter landing page container because those designs depend on a specific width.


src/components/layout/index.js
```diff
@@ -52,9 +51,7 @@ const Layout = ({
       />
       <main id="main">
         <SkipNavContent />
-        <Container narrow={narrow} centered={centered}>
-          {children}
-        </Container>
+        <Container centered={centered}>{children}</Container>
```
schwartzadev, line 54: Can we just use `centered` here instead of `centered={centered}`?

src/stories/2-components/01-Layout.stories.js
```diff
@@ -36,7 +36,21 @@ containerNarrow.story = {
...
+export const containerCentered = () => (
+  <Container centered>
+    <SampleContainer />
+  </Container>
+)
+containerCentered.story = {
+  parameters: {
+    info: {
+      text:
+        'A contered container, the default container for long-form content..',
```
schwartzadev, line 53: * See Typo CI
* Can you note that the centered container also uses the narrow styles?

src/components/layout/index.js
kevee: Nope! It's passing a prop down from `<Layout centered>` down to `Container`

src/stories/2-components/01-Layout.stories.js
kevee: Typo CI finally caught something! I'll fix this.

kevee:
Note that these three PRs are going to be merged all together: #1130, #1103, #1109

PR #1133, opened 2020-06-29
Tfs/form design

tiagofsanchez:
Figma file [here](https://www.figma.com/file/GwkYSPVlTkdgLyGNgeMrLz/v3-Refresh?node-id=177%3A3745)

Changes to to styling according to the design above. 
This PR is not final as we need to 

- [x] get involved from not "budding up" properly, review our Grid design system

- [x] The detail description for any input is on the below the input box do we prefer that to be on the top of the input box? my personal preference in below the input field


src/components/common/form.js
```diff
@@ -58,6 +57,40 @@ const Textarea = ({
...
+        {label}
+      </FormLabel>
+      <input
+        name={name}
+        id={id}
+        type={type}
+        aria-required={isRequired}
+        onChange={onChange}
+        maxLength={maxLength}
+        className={`${className} ${
```
kevee, line 83: I'm seeing a lot of inputs with class of `undefined null` - Let's import the `classnames` module and do:

```
className={classnames(className, detailText ? formStyles.inputDescription : false)}
```

The classnames module will just skip empty values and should return a cleaner class name.

src/components/common/form.js
```diff
@@ -105,15 +142,19 @@ const Select = ({
       name={name}
       onChange={onChange}
       value={value}
-      className={formStyles.select}
+      className={`${formStyles.select} ${
```
kevee, line 145: Ditto for class names here

src/components/common/form.js
```diff
@@ -123,7 +164,11 @@ const List = ({ type, name, options, label, isRequired, detailText }) => (
       <FormLabel isRequired={isRequired}>
         <legend>{label}</legend>
       </FormLabel>
-      <div className={formStyles.fieldsetOptions}>
+      <div
+        className={`${formStyles.fieldsetOptions} ${
```
kevee, line 168: Ditto on class names here.

kevee:
... and yes, let's leave the detail description below the input fields.

PR #1140, opened 2020-06-29
chore(words): ARPC is becoming the Center for Antiracist Research

goldfarb:
The Antiracist Research & Policy Center is becoming the Center for Antiracist Research - this is updating the text that is in /src files and not Contentful.

~~I did not update all of the __test__ files, should those be updated as well (they seem to have out of date language in a couple of other places, which is why I left them as they are - let me know if they should be updated as well~~

UPDATE: __ test __ files have been updated


kevee:
Thanks, @goldfarb! Looks like the tests did fail, so if you can run `npm run test:update` and commit any changes we should be good to go!

PR #1156, opened 2020-07-01
Tfs/blog pagination

tiagofsanchez:
First stab at blog pagination (only for the `/blog` and not the categories) 

To check: 
- Functionality
- Styling and design of the navigation menu


src/templates/blog-list-template.js
```diff
@@ -2,16 +2,26 @@ import React from 'react'
 import { graphql } from 'gatsby'
 import Layout from '~components/layout'
 import BlogTeaserList from '~components/pages/blog/blog-teaser-list'
+import BlogNavigation from '~components/pages/blog/blog-pagination'
```
schwartzadev, line 5: Can we keep either the `BlogNavigation` or the `BlogPagination` name, but not both?
Let's either:
1. Rename this import to BlogPagination
2. Rename the BlogNavigation js and scss module files

src/components/pages/blog/blog-pagination.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react'
+import { Link } from 'gatsby'
+import blogPaginationStyle from './blog-pagination.module.scss'
+
+export default ({ currentPage, numPages }) => {
+  const isFirst = currentPage === 1
+  const isLast = currentPage === numPages
+  const prevPage =
+    currentPage - 1 === 1 ? '/blog' : `/blog/${(currentPage - 1).toString()}`
```
schwartzadev, line 9: Let's update the page path here, too

src/components/pages/blog/blog-pagination.js
```diff
@@ -0,0 +1,56 @@
...
+      <div className={blogPaginationStyle.pageNavigation}>
+        <p>Page:</p>
+        <ul className={blogPaginationStyle.pages}>
+          {Array.from({ length: numPages }, (_, i) => (
+            <li
+              key={`pagination-number${i + 1}`}
+              className={blogPaginationStyle.page}
+            >
+              <Link
+                to={`/blog/${i === 0 ? '' : i + 1}`}
```
schwartzadev, line 43: update path here, too

src/components/pages/blog/blog-pagination.module.scss
```diff
@@ -0,0 +1,62 @@
+ul {
```
schwartzadev, line 1: Can you please run prettier on this file? It looks like some of the spacing is off

src/components/pages/blog/blog-pagination.js
```diff
@@ -0,0 +1,56 @@
...
+                className={`${blogPaginationStyle.pageLink} ${
+                  i + 1 === currentPage ? blogPaginationStyle.pageActive : ''
```
schwartzadev, line 45: Can we also conditionally add `aria-current` to the current page? ([Explanation here](https://tink.uk/using-the-aria-current-attribute/))

schwartzadev:
I missed this in the review, but can we also update the page titles to include the current page:

Page one: `Blog`
Page two: `Blog – Page 2`
etc.

gatsby-node.js
```diff
@@ -82,6 +82,22 @@ exports.createPages = async ({ graphql, actions }) => {
     }
   `)
 
+  const posts = result.data.allContentfulBlogPost.nodes
+  const postsPerPage = 6
```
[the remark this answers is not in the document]
tiagofsanchez: No decision yet, i just venture with 6 :)

src/templates/blog-list-template.js
tiagofsanchez: good catch... thank you for that!

src/components/pages/blog/blog-pagination.js
tiagofsanchez: @schwartzadev do check it out if the implementation makes sense

schwartzadev: looks perfect!

PR #1167, opened 2020-07-01
Fix state notes padding

### LINKED ISSUE #1159: Fix indentation on the state note in mobile
The state notes in `/data` and `/data/state/[name]` pages have padding on mobile breakpoints, let's set their `paddingLeft` as zero.

image

schwartzadev:
- breaks state notes out into a separate component
- fixes #1159


src/components/pages/state/state-notes.module.scss
```diff
@@ -0,0 +1,3 @@
+.container {
+  margin-left: 0;
```
kevee, line 2: Let's change this to `@include col(4 6 8, 0 0 0, 0 0 0);`

schwartzadev:
@kevee I reset the Netlify preview, which seems to have worked -- the `state-notes-module--container` class wasn't added in the previous preview. Can you take another look?

PR #1180, opened 2020-07-03
feature(summary-charts): Add toggle to enable viewing full range of data

goleary:
Desktop:
image
Mobile:
image


goleary:
before:
image

after;
image

PR #1188, opened 2020-07-05
Initial update national chart with bar graph and new data

JasnaMRB:
Based on this mock:
image

I'm assuming this replaces the current donut:
image

If so, whoever continues this work can possibly delete the styling in `national-chart.module.scss#donut-chart`. 

Currently updated data is hardcoded. This should be removed with a real query call. 

There should also be an addition of the exact numbers at the end of each bar, and any a11y tags if needed. 

- [x] Replace hardcoded data with GraphQL call when data is available
- [x] Clean up unused donut chart code if applicable
- [x] Add any needed a11y tags
- [x] Double-check that I didn't break the `HorizontalBarChart` component for other consumer (`_StateComulativeDeathsContainer`)
- [x] Update `horizontal-bar-chart` test
- [x] General cleanup

Curent preview:
preview-tracker


schwartzadev:
Closing in favor of #1203

PR #1203, opened 2020-07-06
National race bar chart

schwartzadev:
A collaborative coding effort with @JasnaMRB -- replaces #1188 (I had trouble with authentication for pushing to the original branch.)


tealtan:
Some styling notes!

- [ ] Can we make the font size of the bar labels 1rem?
- [ ] Above the 768px breakpoint, can we reduce the spacing between bars to 0.5rem?
- [ ] Also, the border color between sections on this page should be `slate-200`.

Issue #1147, opened 2020-06-29
Add share image generator to site builder


kevee:
In order to facilitate easier sharing of data:

- [x] Add a new type of "share image" component that will only be used server-side. This might require a separate call in `gatsby-node.js` on `onPostRender` - which is fine!
- [x] Use [repng](https://github.com/jxnblk/repng) to render these templates out as PNG files
- [ ] Wrap these in a separate layout component that incorporates our logo and URL
- [ ] Setup a single share image for all the charts on the state pages.

kevee:
Fixed in #1211

PR #1211, opened 2020-07-07
feat(visualization): Render components to png

PhastPhood:
### Description
Quite hacky!

This PR is a proof of concept to address #1147. It adds a React higher order component that can be used to wrap charts. These charts are then rendered as png files using a gatsby plugin to the `public/img` folder (or wherever specified by the plugin options).

### Example:
hospitalizations

### Details
This uses puppeteer (a headless Chromium instance) to render images. This is the same as what repng does, but is more efficient since it doesn't spin up the browser for every rendered image.

The `gatsby-ssr` part of the plugin registers the components that are wrapped with the higher order component. It also pulls the CSS styles from the SSR process. The actual png rendering happens in `onPostBuild` in `gatsby-node`. This is because `gatsby-ssr` doesn't allow async work, but `gatsby-node` does. I'm writing the HTML/CSS data from `gatsby-ssr` to the build cache so that `gatsby-node` can pick it up and render it.

Additional build time is ~1 minute on my macbook

### Still need to do:
* Do some component refactoring so that the title is included
* Create a wrapper layout with the COVID Tracking logo
* Use React context or something similar so that putting a renderedComponent inside another one doesn't rerender the child component
* Render images using multiple pages at once to speed up the build
* ~~Directly use the file system instead of using file-system-cache to reduce build warnings~~


PhastPhood:
Looks like an additional library may be needed to get this to work with docker? https://stackoverflow.com/questions/17355863/cant-find-install-libxtst-so-6

kevee:
Wow, @PhastPhood, this is great work! I'm changing this PR to draft and then testing a few use cases with it locally.

goleary:
Question about this: Is there some reason that PNGs are preferred to SVGs?  Gatsby is already statically generating the SVGs at build time so I wonder if we could just write those to disk instead of spinning up puppeteer to render them in a separate process.  I suppose if we want to include non svg elements (chart title is often an html element) then this method would help, That said we could also easily make the title a text element in the SVG.

plugins/gatsby-render-components/rendered-component.js
```diff
@@ -0,0 +1,32 @@
...
+        width,
+        height,
+        relativePath,
+        filename,
+      })
+    }
+
+    render() {
+      const { renderOptions, ...passThroughProps } = this.props
+      return <Component {...passThroughProps} />
```
kevee, line 27: All of these components are going to *just* be used to render PNGs, and not be same as page components. We should just return `null` here.

src/components/charts/bar-chart.js
```diff
@@ -13,6 +13,7 @@ import chartStyles from './charts.module.scss'
 import styles from './bar-chart.module.scss'
 
 import colors from '~scss/colors.module.scss'
+import { renderedComponent } from '~plugins/gatsby-render-components'
```
kevee, line 16: Let's remove the render stuff from the bar charts - we're going to probably want to build entirely separate parent components that include bar charts within them.

PhastPhood: Done! Makes sense. Do you have an idea of how you want this to be done or how you want the end result to look? Just cuz there may be some extra work needed around props passing while rendering depending on implementation.

plugins/gatsby-render-components/rendered-component.js
PhastPhood: Done

kevee:
I'm going to merge this into `feature/share-cards` and then reopen a PR with the first implementation.

PR #1253, opened 2020-07-12
chore(summary-charts): fix notes alignment

schwartzadev:


goleary:
There is a very narrow range 1080-1090px of screen widths where the note on "Current Hospitalizations" flows onto the next line.

That said I think this is a fair trade off to make and I'm working on updates to the display of annotations that will remove the notes link entirely (should get merged this week).

PR #1257, opened 2020-07-13
chore(blog): use contentful ids for short blog urls

### LINKED ISSUE #1254: Shorten URLs for share button
Our URLs are beautifully descriptive, but long when trying to place in a tweet:
image

Lets:

1. In `gatsby-node` while [creating a blog page](https://github.com/COVID19Tracking/website/blob/master/gatsby-node.js#L169), also use `createRedirect` to make a redirect from `/[contentful_id]` to `/blog/[slug]`
2. Pass that shortened `/[contentful_id]` URL to the share button component

schwartzadev:
Closes #1254


src/templates/blog-post.js
```diff
@@ -59,6 +59,7 @@ export default ({ data, path }) => {
 export const query = graphql`
   query($id: String!, $blogImages: [String]) {
     contentfulBlogPost(id: { eq: $id }) {
+      id
```
kevee, line 62: the `id` is a random string from Gatsby that can change frequently, let's use `contentful_id` which is also shorter!

schwartzadev: Fixed, thanks!

PR #1269, opened 2020-07-15
Highlight state notes dates

schwartzadev:
Highlights date in the notes field to improve readability:

image


schwartzadev:
Agreed about the date formats -- our style guide prefers `Month Day` instead of `mm/dd`. Maybe sometime we could have a refactor project to standardize those labels...

PR #1296, opened 2020-07-24
chore: Create automatically-updated data definitions page

kevee:
Creates a data definitions page that is made of `Data definition` entities in Contentful.


kevee:
Closing as it was rolled into the data feature

PR #1307, opened 2020-07-30
chore(crdt-dashboard): Fix less than zero percent display

### LINKED ISSUE #1305: CRDT Web Dashboard: Don't show 0% in the data tables if the value isn't 0
_Current Behavior:_
For states where the percentage of population is less than 1%, we display that value as 0% on the CRDT dashboard. Alabama is an example of this, screenshot:

image

_New Behavior:_

- When the percentage is less than 1% but greater than 0, we should display <1% on the dashboard
- To be clear, if the value is exactly 0, we should continue showing 0%.

kevee:
- Refactors CRDT dashboard percentage display to only display `0%` if there is actually zero percent, instead of zero after rounding.
- Fixes #1305

### DIFF
--- a/src/__tests__/components/pages/race/dashboard/__snapshots__/breakdown-tables.js.snap
+++ b/src/__tests__/components/pages/race/dashboard/__snapshots__/breakdown-tables.js.snap
@@ -132,7 +132,7 @@ exports[`Components : Pages : Race : Dashboard : Ethnicity Table renders correct
         <td
           className="borderLeft"
         >
-          0
+          &lt;1
           %
         </td>
       </tr>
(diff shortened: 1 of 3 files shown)


neeshabeee:
Zero percent display issue lgtm!

Could we also piggyback off this PR to update a couple footnotes? Aviva found these while QA'ing.

- [ ] For Guam: Consolidate the two existing footnotes into a single footnote. The text of the new footnote should say: "This data should not be compared with percentage of the the population. Census does not provide race or ethnicity information for Guam."

- [ ] For Wyoming: Consolidate the three existing footnotes into a single footnote. The text of the new footnote should say: "This data should not be compared with percentage of the the population. Wyoming reports non-mutually-exclusive categories for race and ethnicity data. Percentages may sum to more than 100%."

kevee:
Hm, all those footnotes are handled through the publishing spreadsheet itself, @goldfarb can these be merged in the sheet?

PR #1339, opened 2020-08-14
Update download.js

tomByrer:
To me "Download all available data for a single state" button indicates that button itself is a download link, not an accordion button.  The skinny down arrow at the end looks too much like a line or a letter to be much help (atleast in Win10).

Hopefully "Open listing of individual states' data to download" is more clear to most people.

Thanks for all your hard work!


tomByrer:
image

Still somewhat confusing IMHO since:
- "Download all available data for all states" is non-interactive, but the `daily.csv` is.
- "Open listing of individual states' data to download" **is** interactive, but does not look like a button.  But at least the copy is slightly improved.... might require another PR to button-ize the accordion opener.

PR #1349, opened 2020-08-19
feat: Minify class names

kevee:
- Minify class names
- Merge on August 31


kevee:
@schwartzadev this cuts back on some of our overal HTML weight, but I wanted to have another set of eyes on it to make sure all the features still work.

PR #1393, opened 2020-08-31
chore: Require fields in volunteer form

### LINKED ISSUE #1385: "Required" not being enforced for URL and other fields on volunteer form
**Describe the bug**
Fields that are supposed to be required on the volunteer form are not actually required: submissions come in with those fields blank. URL is certainly one such field. All fields should be required except "Referred by."

kevee:
- Fixes #1385 
- Adds state to ensure required fields are entered before unlocking submit button


kevee:
@amandafrench please review the preview site. Thanks!

amandafrench:
I feel so guilty you are doing all this: I am going to scramble you some volunteers ASAP.

Looks good: have tested, and submit button is grayed out until all required fields have content. BUT two changes, please:

1) Please make the penultimate question "What kinds of projects or tasks would you most like to help with" required as well 
2) Can you validate the input in the URL field to make sure it's really a URL? I'm a little uncertain as to whether we should, but I think on balance we should. 

I am probably going to make a few text changes and changes that will require new Zaps soon, but I will do those in a separate issue.

PR #1396, opened 2020-09-01
Reduce blog content width

schwartzadev:
This decreases the line length for the blog.

(This is based on a conversation @kevee and I had a _while_ back, but better late than never!)

* On desktop, the blog content is now 6 cols wide, which is ~65-70 characters per line
* On tablets, the blog content is no longer full width, and is slightly narrower (mileage will vary based on screen width, obviously)


kevee:
I asked around and I think we should go with 65 instead of 60 characters wide.

schwartzadev:
Great! I think this PR puts the line length right around there. Here's an example:

```
To put the above numbers in context, in 2016 the US Department of <-- 66 chars
Health and Human Services estimated the number of residents of <-- 63 chars
nursing homes and residential care facilities to be about 2.1 million, less <-- 76 chars
than 1 percent of the U.S. population. Using this estimate together <-- 68 chars
with our tallies of deaths in LTCs, we can estimate that more than 1 <-- 69 chars
in 28 LTC residents in the US have died of COVID-19 since March, <-- 65 chars
```

From this sample, the average line length ends up being ~68 chars per line. I think that's about as close as we're going to get to 65 with the 12 column layout. Is that OK by you @kevee?

kevee (approved):
Looks good!

PR #1413, opened 2020-09-08
fix(nav): SSR is rendering "All Posts" in navigation instead of "Blog"

### LINKED ISSUE #1409: Prevent "All posts" from appearing instead of blog
The raw HTML output of our navigation uses "All posts" instead of "Blog" which gets swapped by Gatsby on JS load. Let's chase down why it's being swapped out.

tyler-paulson:
Resolves #1409

I looked into this and the string "All Posts" is coming from `blog-category.js` when the return link is passed into the `<Layout />` component.

To fix, I updated `return-links.js` to access the elements of the `topNavigation` array by clone instead of by reference.

I'm still learning how Gatsby's SSR works under the hood but I'm speculating this only occurred in the SSR because `BlogCategoryTemplate` gets imported as its rendering all the pages and not when the JS on the individual pages loads up.

### DIFF
--- a/src/components/layout/header/return-links.js
+++ b/src/components/layout/header/return-links.js
@@ -47,7 +47,7 @@ const ReturnLinks = ({ links, pathNavigation, topNavigation, children }) => {
 
     if (matchingTopNavLink !== -1) {
       // pulls the title info for nav elements
-      returnLinks[index] = topNavigation[matchingTopNavLink]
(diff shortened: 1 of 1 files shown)


kevee (approved):
@tyler-paulson Ha, I was just opening a PR with this same fix. Thanks!

tyler-paulson:
@kevee Ha, great minds think alike! It was a bit of a tricky one but definitely helped me start to figure out Gatsby.

PR #1426, opened 2020-09-12
feat(social-sharing): update hover/focus styles

### LINKED ISSUE #1408: Add social sharing hover style
How about we add a dashed/dotted outline to the share icons when they're hovered.

schwartzadev:
Fixes #1408


schwartzadev:
Merge at will @kevee!

PR #1454, opened 2020-09-22
fix(visualization): Exclude negative numbers from the 7-day average p…

### LINKED ISSUE #1415: Proposal: Exclude negative numbers from the 7-day average plot
Large outlier negative numbers skew the 7-day average line significantly (and, arguably, in a way that does not benefit the reader). May I propose that such that extreme values be excluded from the plot?

e.g. North Carolina
92125652-42062900-edcd-11ea-8149-fd930bcedee2

AaronMullan:
…lot Fixes #1415


kevee (approved):
Thanks so much, @AaronMullan!

PR #1458, opened 2020-09-23
fix(format-number): round to one decimal

### LINKED ISSUE #1455: crdt racial dash multiplier rounding
Right now, the headline for the viz goes out to four sf. Let's reduce that to just one decimal
image

schwartzadev:
Fixes #1455


src/components/utils/format.js
```diff
@@ -31,9 +31,11 @@ const FormatDate = ({
   )
 }
 
-const FormatNumber = ({ number, nullValue = false }) => (
-  <>{number || number === 0 ? number.toLocaleString() : nullValue || 'N/A'}</>
-)
+const FormatNumber = ({ number, nullValue = false }) => {
```
kevee, line 34: @schwartzadev what if we make the option for precision a prop for `FormatNumber`?

kevee (changes requested):
I think we might want to make precision a prop so we can re-use this feature and also existing implementations won't get rounded.

src/components/utils/format.js
schwartzadev: Good call!

kevee (approved):
LGTM! Thanks!

PR #1549, opened 2020-10-30
fix(a11y): Use image description in blog posts, or set to aria-hidden

kevee:
- Fixes confusing alt text in blog image blocks


schwartzadev (approved):
Looks great, thanks!

Issue #1608, opened 2020-11-19
Add missing space for civil services citation


schwartzadev:
Looks like we're missing a space between `Source:` and `Civil Services`...let's add it in there!
![image](https://user-images.githubusercontent.com/18607205/99702179-b8bcd500-2a52-11eb-9e9f-1e5f88f64c1b.png)

Edit: On this page: https://covidtracking.com/race/get-better-data

ivarvong:
There seems to be an (unlikely!) collision on the `._9c10` selector.

`.toggle-wrapper` [0] is the first rule to output to this selector, `.detail-text` [1] is the second.

The `display: flex` from [0] causes the existing space [2] to be collapsed for "Source: Civil Services". 

To confirm this, add something like `--mark-conflict: 1` to `.toggle-wrapper` and inspect the computed CSS for the component [2].

The behavior of the component changes when this is fixed, specifically that `text-align: right` [3] will apply above `$viewport-sm`. I'm not sure if that's desirable.

Locally, changing the `cssLoaderOptions` hash length from 4 to 5 fixes the selector collision. [4]

0: https://github.com/COVID19Tracking/website/blob/5a814c7998e7c8dca108a48c3943b10e40acbb58/src/components/common/toggle.module.scss#L24-L27
1: https://github.com/COVID19Tracking/website/blob/3dd92ac6fdb0e4a631d999a7b6242783d78431f5/src/components/common/detail-text.module.scss#L1-L13
2: https://github.com/COVID19Tracking/website/blob/3dd92ac6fdb0e4a631d999a7b6242783d78431f5/src/components/pages/race/get-better-data/state-contact.js#L61
3: https://github.com/COVID19Tracking/website/blob/3dd92ac6fdb0e4a631d999a7b6242783d78431f5/src/components/pages/race/get-better-data/state-contact.module.scss#L2
4: https://github.com/COVID19Tracking/website/blob/master/gatsby-config.js#L62

theroyalpekka:
Am a beginner, able to resolve the problem by changing the styling of the page , is this a good way to solve this issue ?

schwartzadev:
Closed by #1633

PR #1633, opened 2020-11-24
fix(design): Remove style collision removing space

JackHowa:
- Test and didn't see updated snapshots 
- Noticed the react components did have the child shown of the space, but, due to collisions, wasn't rendering 
-> investigation here showed me https://github.com/COVID19Tracking/website/issues/1608#issuecomment-731599590 (thanks @ivarvong )

- see source space between `:` and `Civil services`

before: 

Screen Shot 2020-11-23 at 6 20 01 PM
Screen Shot 2020-11-23 at 6 20 35 PM

after: 

Screen Shot 2020-11-23 at 6 19 18 PM
Screen Shot 2020-11-23 at 6 18 41 PM

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -59,7 +59,7 @@ const gatsbyConfig = {
       options: {
         data: sassImports,
         cssLoaderOptions: {
-          localIdentName: '[sha1:hash:hex:4]',
+          localIdentName: '[sha1:hash:hex:5]',
         },
       },
     },
--- a/src/components/common/toggle.module.scss
+++ b/src/components/common/toggle.module.scss
@@ -22,7 +22,6 @@
 }
 
 .toggle-wrapper {
-  display: flex;
   align-items: center;
 }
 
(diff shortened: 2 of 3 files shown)


ivarvong:
Does removing "display: flex" affect the behavior of the component(s) using it? (I haven't checked)

If it's feasible, fixing the selector collision seems desirable.

JackHowa:
yeah true, the underlying issue is a config

JackHowa:
> Does removing "display: flex" affect the behavior of the component(s) using it? (I haven't checked)

From playing around with clicking the button and hovering vs prod, I haven't noticed a difference

JackHowa:
huh updating to length did not fix the issue 

Screen Shot 2020-11-23 at 7 04 38 PM

via https://github.com/webpack/loader-utils#interpolatename

JackHowa:
yep updated ... added back the `display: flex` with the longer longer.

JackHowa:
@ivarvong as you alluded to, whether we want to still align right for the source is open. Seems like that was accidentally shown on the left side previously for all breakpoints

schwartzadev (approved):
Awesome, looks good to me! Thank you both @JackHowa + @ivarvong

PR #1640, opened 2020-11-29
fix(contact): Edit id and name values for form on data contact page

### LINKED ISSUE #1638: Rename id for Data Issue contact form date field
This is a tiny unimportant issue, maybe good first task for a new person, but on https://github.com/COVID19Tracking/website/blob/master/src/pages/contact/data.js can the date id be changed from "contact-date" to "contact-issue-start-date" or something similar? The label `contact-date` that comes in makes it look like it's the date the user contacted us instead of the user-reported "date when the data issue began." I could do it but I don't know if that id is being used elsewhere in the code. 

```
<Input
          type="date"
          label="Date when issue began"
          id="contact-date"
          name="contact-date"
          isRequired
          onChange={event => setDate(event.target.value)}
        />
```

ezraschrage:
fixes #1638
The name and id for contact-date appear to only have relevance for the webhook the form is submitting to. It does not appear to affect anything else in the website code, but I cannot say if it is being used in whatever is receiving the JSON. I hope this was useful!

### DIFF
--- a/src/pages/contact/data.js
+++ b/src/pages/contact/data.js
@@ -47,8 +47,8 @@ const ContactPage = ({ data }) => {
         <Input
           type="date"
           label="Date when issue began"
-          id="contact-date"
-          name="contact-date"
+          id="contact-issue-start-date"
+          name="contact-issue-start-date"
           isRequired
           onChange={event => setDate(event.target.value)}
         />


schwartzadev (approved):
Looks good, thanks @ezraschrage

PR #1646, opened 2020-11-30
feat: Enable data warning box through contentful

kevee:
Allows setting a data warning box using a Contentful snippet.


kevee:
Todo:
- [ ] Fix the test snapshot mocks

PR #1647, opened 2020-12-01
 feat: Pre-populate states on "Get better data" form.

### LINKED ISSUE #1643: Pre-populate states on "Get better data" form
### Goal

* When users click the `Help us get better data` link in the data page empty state, pre-populate that state on the dropdown

### Implementation

* Use the style: `/race/get-better-data#NY`
* Emulate the URL parsing on the `/race/infection-and-mortality-data#NY` page
* Update the `/data` page links to point to include the state anchor as appropriate

rubenschuckit:
This change resolves #1643


schwartzadev (changes requested):
This looks great, thank you so much @rubenschuckit!

There are two things that I'd love to change, though:
## Scroll position

### Expected behavior
1. Navigate to data page on the preview site (i.e. [Guam](https://deploy-preview-1647--upbeat-lovelace-3e9fff.netlify.app/data#state-gu))
2. Click on `Help us get better data`
3. Navigate to the top of the [Get better data page](https://deploy-preview-1647--upbeat-lovelace-3e9fff.netlify.app/race/get-better-data#GU)

### Actual behavior
* 1-2 as expected
* Clicking on the link navigates to the bottom of the `Get better data` page.

I think this is something from Gatbsy: perhaps it is retaining the very scrolled down position on the data page when users click over to the `Get better data` page. @kevee is this something you've encountered before with Gatsby?



## State hash is fixed

### Expected behavior
* Changing the `select` selection updates the URL hash (same as https://covidtracking.com/race/infection-and-mortality-data#NV)

### Actual behavior
* The URL hash does not change when the selection changes

It would be amazing if you could update this behavior to be consistent with the [existing pattern](https://covidtracking.com/race/infection-and-mortality-data#NV).


Thanks again @rubenschuckit 🎉

rubenschuckit:
Thanks for the detailed comments. 

I've gone ahead and added the URL behavior when selecting a state from the dropdown. I'll look at the scroll issue this evening. 

One thing to note is when you change the URL hash by selecting a new state it adds to the history. So in the [example](https://covidtracking.com/race/infection-and-mortality-data#NV) you shared, if you select multiple states, when you try to go back it cycles through the hashes (and since nothing is listening to the URL the dropdown doesn't change).

Do you want me to file an issue for that? What is expect behavior?

schwartzadev:
@rubenschuckit Great call for the history. Thanks so much for taking a look.

> So in the example you shared, if you select multiple states, when you try to go back it cycles through the hashes (and since nothing is listening to the URL the dropdown doesn't change).

> What is expect behavior?

Assuming this pattern:

  * Navigated to `/data/`
  * Navigated to `/race/infection-and-mortality-data#NV`
  * Changed the select option to `/race/infection-and-mortality-data#NM`
  * Pressed the back button

... I think users would either expect to either:
  * Return to the `NV` card OR
  * Return to the `/data/` page

Considering that we're already updating the URLs, let's do the former option. That is, listen to the URL and update state if the URL changes.

> Do you want me to file an issue for that?
If you want to take this on yourself, by all means open a PR and tackle it. If not, please open an issue.

Thanks again!

rubenschuckit:
Sounds good. I'll add that to this PR. In the meantime, any guidance on the scroll issue? I'm new to React/Gatsby and didn't find anything in a quick search. We could use JS to scroll to the top, but that would be pretty hacky.

rubenschuckit:
I didn't create a new branch in my fork, which created a git catastrophe when trying to create a PR into the same branch upstream. Remaking...

PR #1654, opened 2020-12-03
chore(get-involved): add name fields

### LINKED ISSUE #1589: Remove firstname, lastname, and zip code from newsletter page
Remove fields for firstname, lastname, and zip code from newsletter signup form at /about/newsletter

ezraschrage:
fixes #1589
Added name fields back to the get involved page. I used the Col component to organize the input fields into one column since three fields are better organized as a single column rather than the original four in a 2x2 grid. Attached is a screenshot of the final look. It looks similar if implemented with the Row component or without a grid if that's a preferable frame for it to be in, so I can change it to those if need be.

Newsletter with name fields

### DIFF
--- a/src/components/pages/get-involved/form.js
+++ b/src/components/pages/get-involved/form.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Row } from '~components/common/grid'
+import { Col } from '~components/common/grid'
 import { Form, Input } from '~components/common/form'
 import formStyle from './form.module.scss'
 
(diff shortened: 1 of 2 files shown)


src/components/pages/get-involved/form.js
```diff
@@ -10,18 +10,29 @@ const GetInvolvedForm = () => (
...
-    </Row>
-
+    <Col width={[4, 6, 6]} padding={[0, 0, 0]} className={formStyle.grid}>
+      <Input
+        label="First name"
+        type="text"
+        name="FNAME"
+        id="action-form-first-name"
+        isRequired
+      />
```
schwartzadev, line 20: These look great, thanks for copying them over!

src/components/pages/get-involved/form.js
```diff
@@ -10,18 +10,29 @@ const GetInvolvedForm = () => (
...
-          label="Email address"
-          type="email"
-          name="EMAIL"
-          id="action-form-email"
-          isRequired
-        />
-      </div>
-    </Row>
-
+    <Col width={[4, 6, 6]} padding={[0, 0, 0]} className={formStyle.grid}>
```
schwartzadev, line 13: Could you please center this on Desktop?
image

schwartzadev (changes requested):
Looking awesome, thanks @ezraschrage!

Just one thing to 🔧 fix up and this should be good to go.

Thank you so much.

src/components/pages/get-involved/form.js
ezraschrage: Do you mean to extend the input bars to the width of the column like this?
Screen Shot 2020-12-05 at 8 31 24 PM

Presently it is centered like the other content on the website.
combo

schwartzadev: Yes! That would follow the pattern for our [volunteer form](https://covidtracking.com/contact/volunteer). Sorry that was unclear, thanks again!

ezraschrage: No worries! Happy to help out.

schwartzadev (approved):
Awesome, looks perfect! Thank you @ezraschrage

PR #1656, opened 2020-12-06
feat: Listen to URL hash change in race/infection-and-mortality-data to update state.

rubenschuckit:
When selecting a new state in the dropdown in `/race/infection-and-mortality-data` it adds the URL with the new hash to history. Previously when navigating back after selecting two or more states, the hash would change but the state would not update. Now we listen for the hash change and update the state with backwards navigation.


schwartzadev (approved):
LGTM, thanks!

PR #1650, opened 2020-12-02
feat(definitions-panel): style scrollbar

schwartzadev:
This adds custom styles to the scrollbar, which makes
the clash in #1644 less prominent.

Note that this obscures the scrollbar buttons.
(See [this Stack Overflow post](https://stackoverflow.com/questions/47576815/how-to-add-arrows-with-webkit-scrollbar-button))

@kevee how do you feel about the buttons from an a11y standpoint?

PR #1664, opened 2020-12-10
fix(design): Remove border from dialog to ensure full-height scrollbar 

### LINKED ISSUE #1644: Fix definitions sidebar color/alignment
On Chrome, we have a little bit of purple peeking out behind the definitions sidebar scrollbar:

image

Let's style the scrollbar to have the same `#111354` background so this small corner isn't appearing on its own. Since this is not a Firefox issue, we can use the `::-webkit-scrollbar` psuedo element.

JackHowa:
- Ran `yarn test` to ensure not breaking any snapshots 

Closes #1644 
Closes #1650

On page https://covidtracking.com/data, then clicking the little Definition link, it looks like removing the 2rem border ensures the nested scrollbar is full-height (no purple below). 

If we want, I can try to look into replicating the design with the full-height scrollbar. I haven't done cross-browser testing on this but it seems to be working as expected with chrome. 

definition link 
Screen Shot 2020-12-10 at 12 26 06

before: 

Screen Shot 2020-12-10 at 12 28 35

after: 

Screen Shot 2020-12-10 at 12 30 43


JackHowa:
huh, it looks like this top part of the text is getting caught  at the top. the top area is covering the starting text on click 

Screen Shot 2020-12-10 at 17 14 00

schwartzadev (approved):
Perfect, thanks! @JackHowa

PR #1686, opened 2020-12-18
fix (screenshots): state variable for getting taco YAML configs, currently hardcoded to CA

julia326:


kevee (approved):
Thanks, @julia326!

PR #1752, opened 2021-01-16
fix: national forEach annotation bug

schwartzadev:
Handle cases where annotations = false before attempting
to iterate over the annotations.

Also updates the national definition panel title to say
'National' instead of 'undefined'.


kevee:
@schwartzadev I think this is fine but Netlify fell over so I kicked off another preview build.

kevee (approved):
LGTM!

PR #1880, opened 2021-05-18
chore: Add preserve thread function

kevee:
Adds webhook to store slack threads


kevee:
Closing, we are moving this to its own repo

PR #1894, opened 2022-02-08
Remove proxy, security bump and test snapshot

### LINKED ISSUE #1889: failing tests due to hardcoded year in expected results
Tests are failing because [the year is hardcoded into the expected results](https://github.com/COVID19Tracking/website/blob/bdee11f90f18cedc1732e10927a5ff585bffec85/src/__tests__/components/layout/__snapshots__/footer.js.snap#L132).

This seems to be a `jest` thing and I don't know if it's supposed to get automatically updated or if there's a way to make that particular part of the results more forgiving so that you don't have to update it every year, or what, but taking a moment to log the issue. Thanks.

kevee:
- Removes proxy for Screenshots
- Removes API redirects for past versions
- Bump nanoid. Fixes #1893 
- Re-run jest snapshots. Fixes #1889


kevee:
Merging, ignoring CircleCI since we have now removed it.

